PR #12, opened 2025-02-17
Add documentation about configuring SLURM on GH200

msimberg:
This is a starting point for docs about configuring slurm for GH200. It covers single and multiple ranks per GPU. It mentions the default process mode which can be a big footgun, and recommends use of MPS. The MPS wrapper script is the one from the current knowledge base.

I expect some refactoring might be useful if/when there are MI300 docs for slurm as some of it might be similar, but I haven't attempted to put that in a generic section yet at this point.


docs/tools/slurm.md
```diff
@@ -20,37 +20,103 @@ The following sections will provide detailed guidance on how to use SLURM to req
 [](){#gh200-slurm}
 ### NVIDIA GH200 GPU Nodes
 
-!!! todo
-    document how slurm can be used on the Grace-Hopper nodes.
+The [GH200 nodes on Alps][gh200-node] have four GPUs per node, and SLURM job submissions must be configured appropriately to best make use of the resources. Applications that can saturate the GPUs with a single process per GPU should generally prefer this mode. [Configuring SLURM jobs to use a single GPU per rank][gh200-slurm-single-rank-per-gpu] is also the most straightforward setup. Some applications perform badly with a single rank per GPU, and require use of [NVIDIA's Multi-Process-Service (MPS)](https://docs.nvidia.com/deploy/mps/index.html) to oversubscribe GPUs with multiple ranks per GPU.
 
-    Note how you can link to this section from elsewhere using the anchor above, e.g.:
+The best SLURM configuration is application- and workload-specific, so it is worth testing which works best in your particular case. Also see [TODO][TODO] for information about recommended application-specific SLURM configurations.
```
msimberg, line 25: This is still to do. I'd like to link to the "Scientific applications" section here (which does not yet exist).

docs/tools/slurm.md
```diff
@@ -20,37 +20,103 @@ The following sections will provide detailed guidance on how to use SLURM to req
 [](){#gh200-slurm}
 ### NVIDIA GH200 GPU Nodes
 
-!!! todo
-    document how slurm can be used on the Grace-Hopper nodes.
+The [GH200 nodes on Alps][gh200-node] have four GPUs per node, and SLURM job submissions must be configured appropriately to best make use of the resources. Applications that can saturate the GPUs with a single process per GPU should generally prefer this mode. [Configuring SLURM jobs to use a single GPU per rank][gh200-slurm-single-rank-per-gpu] is also the most straightforward setup. Some applications perform badly with a single rank per GPU, and require use of [NVIDIA's Multi-Process-Service (MPS)](https://docs.nvidia.com/deploy/mps/index.html) to oversubscribe GPUs with multiple ranks per GPU.
```
bcumming, line 23: I have a habit of putting one sentence per line with no wrapping in markdown
* the sentences will still be rendered as paragraphs
* it often reduces the chance of nasty merges where two people edit the same paragraph.
    * you are still stuffed if they edit the same sentence.

msimberg: Yep, there are downsides to any approach. I've split up the lines in https://github.com/eth-cscs/cscs-docs/pull/12/commits/7922cd8251af49b4a994ec7d8f0191e9c7af6d9d (for the parts that I changed).

docs/tools/slurm.md
msimberg: Added link in https://github.com/eth-cscs/cscs-docs/pull/12/commits/3211a0bbc0803d8e315cfe96bf9f8f04f73ed676.

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
+If the `mps-wrapper.sh` script is in the current working directory, you can then launch jobs using MPS for example as follows:
+
+```bash
+#!/bin/bash
+#SBATCH --job-name=oversubscription-affinity-test
+#SBATCH --nodes=2
+#SBATCH --ntasks-per-node=32
+#SBATCH --cpus-per-task=8
+
+export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK
```
RMeli, line 119: This is dangerous with OpenBLAS. Maybe it is worth an note somewhere (proper markdown `!!!dange` note, or maybe just a comment)?

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
+Applications that can saturate the GPUs with a single process per GPU should generally prefer this mode.
+[Configuring SLURM jobs to use a single GPU per rank][gh200-slurm-single-rank-per-gpu] is also the most straightforward setup.
+Some applications perform badly with a single rank per GPU, and require use of [NVIDIA's Multi-Process-Service (MPS)](https://docs.nvidia.com/deploy/mps/index.html) to oversubscribe GPUs with multiple ranks per GPU.
+
+The best SLURM configuration is application- and workload-specific, so it is worth testing which works best in your particular case.
+Also see [Scientific Applications][sciapps] for information about recommended application-specific SLURM configurations.
+
+!!! warning
+    The GH200 nodes have their GPUs configured in ["default" compute mode](https://docs.nvidia.com/deploy/mps/index.html#gpu-compute-modes).
+    Unlike "exclusive process" mode, "default" mode allows multiple processes to submit work to a single GPU simultaneously.
```
RMeli, line 33: Do we need to make a note that this changed recently on Alps? Maybe not here in the main documentation, but it would be good to have some sort of CHANGELOG for this kind of things.

Maybe it is also worth adding a note on why they are in "default" mode? I think the expectation from users is "exclusive process" (which is mentioned), and it would be good to give a rationale about this choice (i.e. containers not working otherwise).

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
+
+Configuring SLURM to use one GH200 GPU per rank is easiest done using the `--ntasks-per-node=4` and `--gpus-per-task=1` SLURM flags.
+For advanced users, using `--gpus-per-task` is equivalent to setting `CUDA_VISIBLE_DEVICES` to `SLURM_LOCALID`, assuming the job is using four ranks per node.
+The examples below launch jobs on two nodes with four ranks per node using `sbatch` and `srun`:
 
 === "sbatch"
 
     ```bash
     #!/bin/bash
     #SBATCH --job-name=affinity-test
```
RMeli, line 52:
```suggestion
    #SBATCH --job-name="one-task-per-gpu-example"
```
?

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
     ```
 
 === "srun"
 
     ```
-    > srun -n8 -N2 --gpus-per-task=1 affinity
+    srun --nodes=2 --ntasks-per-node=4 --gpus-per-task=1 <application>
     ```
+    
+Omitting the `--gpus-per-task` flag will lead to all ranks on the node using the first GPU.
```
RMeli, line 66:
```suggestion
Omitting the `--gpus-per-task` flag results in `CUDA_VISIBLE_DEVICES` being unset, which will lead to all ranks on the node using the first GPU.
```
This is also something that changed at some point IIRC, where before all GPUs were visible with `CUDA_VISIBLE_DEVICES=0,1,2,3`?

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
+
+[](){#gh200-slurm-multi-rank-per-gpu}
+#### Multiple ranks per GPU
+
+Using multiple ranks per GPU can improve performance e.g. of applications that don't generate enough work for a GPU using a single rank, or ones that scale badly to all 72 cores of the Grace CPU.
+In these cases SLURM jobs must be configured to assign multiple ranks to a single GPU.
+This is best done using MPS.
```
RMeli, line 73:
```suggestion
This is best done using [NVIDIA's Multi-Process-Service (MPS)](https://docs.nvidia.com/deploy/mps/index.html).
```
Use the same link as above, in case someone jumps directly to this section?

docs/tools/slurm.md
```diff
@@ -20,37 +20,116 @@ The following sections will provide detailed guidance on how to use SLURM to req
...
+
+exit $result
+```
+
+Save the above script as `mps-wrapper.sh` and make it executable with `chmod +x mps-wrapper.sh`.
+If the `mps-wrapper.sh` script is in the current working directory, you can then launch jobs using MPS for example as follows:
+
+```bash
+#!/bin/bash
+#SBATCH --job-name=oversubscription-affinity-test
```
RMeli, line 114:
```suggestion
#SBATCH --job-name="mps-example"
```

docs/tools/slurm.md
msimberg: Yeah, good idea. I updated this and the other example to `gh200-single-rank-per-gpu` and `gh200-multiple-ranks-per-gpu` to be very explicit. Do you think that's also good?

docs/tools/slurm.md
msimberg: > This is also something that changed at some point IIRC, where before all GPUs were visible with CUDA_VISIBLE_DEVICES=0,1,2,3

I think my wording needs a bit of polishing... I don't think anything changed, it's just that most applications will use CUDA with the device set to 0 even when all devices are exposed. Let me try some clarification here.

RMeli: Sorry, probably my comment wasn't clear but it was related to https://github.com/eth-cscs/cscs-docs/pull/12#discussion_r1959327733, where I was suggesting it would be useful to have a sort of CHANGELOG. I think this changed at some point (not sure if from Todi to Daint or in a Daint updated), but it's un-related to what you wrote (where I just think clarifying that `CUDA_VISIBLE_DEVICES` will remain unset might be useful).

docs/tools/slurm.md
RMeli: Yep, that's great.

docs/tools/slurm.md
msimberg: Too short: https://github.com/eth-cscs/cscs-docs/pull/12/commits/2add4041c818cbf356114fadf3e8b3d9dbac22fb?

docs/tools/slurm.md
RMeli: Maybe a good candidate for a [code annotation](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations)?

docs/tools/slurm.md
msimberg: Ah, I think I know what you're referring to now. The `--gpus-per-task` flag used to simply have no effect at all (or maybe result in jobs failing to start; in any case, it did not set `CUDA_VISIBLE_DEVICES`), but I'm pretty sure that was before users came on the system.

See other comment for the changelog (it exists on the current KB, but hasn't been ported yet).

docs/tools/slurm.md
RMeli: I think it's good enough. If people need/want more information, they can always ask.

docs/tools/slurm.md
msimberg: Very good point. I was unsure about including this here in the first place (mixing GPU bindings and CPU affinity in the same example; I also don't mention e.g. `OMP_NUM_THREADS` in the "single rank per GPU" section at the moment).

I'm thinking about adding a separate section talking about recommended CPU bindings/OMP_NUM_THREADS etc. Do you think that makes sense? If yes, I would make a follow-up to this PR. In https://github.com/eth-cscs/cscs-docs/pull/12/commits/b350064cb59c44cf556c5d29b6169b02112dec1e I've (for now) removed most mentions of CPU bindings.

bcumming (approved):

docs/tools/slurm.md
RMeli: Yes, it makes sense to have this separately.

PR #22, opened 2025-02-24
vscode --name

henrique:
you should set --name on tunnel or it'll use the first node ID but this name is persisted across sessions.
this info is store under ~/.vscode-server (this path can be changed with --cli-data-dir but I am not sure what happens when it gets [partially] deleted by the scratch policy...)

### DIFF
--- a/docs/access/vscode.md
+++ b/docs/access/vscode.md
@@ -57,12 +57,14 @@ run the `code` executable that you downloaded the `tunnel` argument.
 You will be asked to choose whether to log in to Microsoft or GitHub (we have tested with GitHub):
 
 ```
-> code tunnel
+> code tunnel --name=$CLUSTER_NAME-tunnel
 ...
 ? How would you like to log in to Visual Studio Code? ›
   Microsoft Account
 ❯ GitHub Account
 ```
+
+* `--name` sets a name for the tunnel, which will be later listed on the VSCode UI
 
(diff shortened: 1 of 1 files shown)


henrique:
@bcumming by 'Connecting via VSCode UI' do you mean the Remote-SSH?

docs/access/vscode.md
```diff
@@ -57,13 +57,15 @@ run the `code` executable that you downloaded the `tunnel` argument.
 You will be asked to choose whether to log in to Microsoft or GitHub (we have tested with GitHub):
 
 ```
-> code tunnel
+> code tunnel --name=$CLUSTER_NAME-tunnel
```
bcumming, line 60: I noticed this - it was calling all connections `daint-ln002` by default, for example.

docs/access/vscode.md
```diff
@@ -57,13 +57,15 @@ run the `code` executable that you downloaded the `tunnel` argument.
...
 ...
 ? How would you like to log in to Visual Studio Code? ›
   Microsoft Account
 ❯ GitHub Account
 ```
 
+* `--name` sets a name for the tunnel, which will be later listed on the VSCode UI
```
bcumming, line 67: We could use an admonition:
https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:tip
```
!!! tip
    Give the tunnel a unique name using the `--name` flag, which will later be listed on the VSCode UI.
```

bcumming (approved):
lgtm!

PR #36, opened 2025-03-11
Enhance uenv docs

RMeli:
The main change is the position of the note for `uenv repo create`, which users seem to miss because it comes _after_ `uenv image pull`.

Other changes are small an include:
* Removal of outdated warning
* `uenv2` installation instructions from source not in an `example` block but as part of the "Installing the uenv tool" section


bcumming (approved):
thanks!

bcumming (approved):
thanks!

PR #39, opened 2025-03-12
Add internet access page

msimberg:
Ports https://confluence.cscs.ch/spaces/KB/pages/868820917/Internet+access+on+the+Alps+infrastructure to the new docs. Currently under "Guides".

@bcumming do you think it makes sense to rename "Guides" already now or shall we keep it like this for now.

This section could fit close to the "Alps High Speed Network" section, but I think it's a bit too hidden to put it there. Additionally, I think it's good if the hardware is focused on the hardware itself and doesn't go into implementation details like this (i.e. that a proxy server is required for compute nodes).

### DIFF
--- a/docs/guides/internet-access.md
+++ b/docs/guides/internet-access.md
@@ -9,12 +9,12 @@ Login nodes have public IP addresses which means that they can directly access t
     Compute nodes are configured with the following environment variables to use the proxy server:
     
     ```bash
-    https_proxy=http://proxy.cscs.ch:8080
-    http_proxy=http://proxy.cscs.ch:8080
-    no_proxy=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
-    HTTPS_PROXY=http://proxy.cscs.ch:8080
-    HTTP_PROXY=http://proxy.cscs.ch:8080
-    NO_PROXY=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    export https_proxy=http://proxy.cscs.ch:8080
+    export http_proxy=http://proxy.cscs.ch:8080
+    export no_proxy=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    export HTTPS_PROXY=http://proxy.cscs.ch:8080
+    export HTTP_PROXY=http://proxy.cscs.ch:8080
+    export NO_PROXY=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
     ```
 
 !!! warning "Public IPs are a shared resource"
--- a/docs/guides/internet-access.md
+++ b/docs/guides/internet-access.md
@@ -37,7 +37,7 @@ Match Host *,!148.187.0.0/16,!192.168.0.0/16,!172.16.0.0/12,!10.0.0.0/8exec "hos
 
 This configuration takes into account that login and compute nodes require a different setup.
 
-??? info "Error message when cloning without the proxy set up for SSH"
+??? warning "Error message when cloning without the proxy set up for SSH"
     When cloning a git repository without the correct SSH configuration, cloning will time out as follows:
     ```bash
     [daint][<user>@daint-ln001 ~]$ git clone git@github.com:open-mpi/ompi.git
--- a/docs/guides/internet-access.md
+++ b/docs/guides/internet-access.md
@@ -1,0 +1,52 @@
+[](){#ref-guides-internet-access}
+# Internet Access on Alps
+
+The [Alps network][ref-alps-hsn] is mostly configured with private IP addresses (`172.28.0.0/16`).
+Login nodes have public IP addresses which means that they can directly access the internet, while a proxy server provides internet access for compute nodes.
+
+??? info "Compute node proxy configuration"
+
+    Compute nodes are configured with the following environment variables to use the proxy server:
+    
+    ```bash
+    https_proxy=http://proxy.cscs.ch:8080
+    http_proxy=http://proxy.cscs.ch:8080
+    no_proxy=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    HTTPS_PROXY=http://proxy.cscs.ch:8080
+    HTTP_PROXY=http://proxy.cscs.ch:8080
+    NO_PROXY=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    ```
+
+## Using SSH through the proxy server 
+
+While use of the proxy server is transparent for most use cases, e.g. cloning git repositories from GitHub over SSH requires additional configuration for compute nodes.
+Cloning over https works without additional configuration.
+To make SSH use the proxy server, add the following to your `~/.ssh/config` file:
+
+```bash
+Match Host *,!148.187.0.0/16,!192.168.0.0/16,!172.16.0.0/12,!10.0.0.0/8 exec "hostname -I | grep -vqF 148.187."
+    ProxyCommand nc -X connect -x proxy.cscs.ch:8080 %h %p
+```
+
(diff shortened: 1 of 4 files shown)


docs/guides/internet-access.md
```diff
@@ -0,0 +1,52 @@
+[](){#ref-guides-internet-access}
+# Internet Access on Alps
```
msimberg, line 2: Maybe this is sufficient:
```suggestion
# Internet Access
```
?

docs/guides/internet-access.md
```diff
@@ -0,0 +1,52 @@
...
+    http_proxy=http://proxy.cscs.ch:8080
+    no_proxy=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    HTTPS_PROXY=http://proxy.cscs.ch:8080
+    HTTP_PROXY=http://proxy.cscs.ch:8080
+    NO_PROXY=.local, .cscs.ch, localhost, 148.187.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+    ```
+
+## Using SSH through the proxy server 
+
+While use of the proxy server is transparent for most use cases, e.g. cloning git repositories from GitHub over SSH requires additional configuration for compute nodes.
```
RMeli, line 22:
```suggestion
While use of the proxy server is transparent for most use cases, 
others need additional configuration for compute nodes.
An example is cloning git repositories from GitHub over SSH.
```
Not sure about the phrasing starting with "e.g.".

docs/guides/internet-access.md
```diff
@@ -0,0 +1,52 @@
...
+To make SSH use the proxy server, add the following to your `~/.ssh/config` file:
+
+```bash
+Match Host *,!148.187.0.0/16,!192.168.0.0/16,!172.16.0.0/12,!10.0.0.0/8 exec "hostname -I | grep -vqF 148.187."
+    ProxyCommand nc -X connect -x proxy.cscs.ch:8080 %h %p
+```
+
+This configuration takes into account that login and compute nodes require a different setup.
+
+!!! info "Error message when cloning without the proxy set up for SSH"
```
RMeli, line 33:
```suggestion
??? info "Error message when cloning without the proxy set up for SSH"
```
Maybe this can be collapsed? I'm assuming here that collapsed blocks do show up in a search.

docs/guides/internet-access.md
```diff
@@ -0,0 +1,52 @@
...
+    ssh: connect to host github.com port 22: Connection timed out
+    fatal: Could not read from remote repository.
+
+    Please make sure you have the correct access rights
+    and the repository exists.
+    ```
+
+## Accessing the public IP of a node
+
+When on a login node configured with a public IP address, you can retrieve the public IP address for example as follows:
```
RMeli, line 47: Maybe it is worth adding an info/warning somewhere pointing out that the public IP address is essentially a shared resource, and therefore limitations based on the IP address (such as pulls from DockerHub) are shared across all users?

msimberg: Good idea!

msimberg: I added a note in https://github.com/eth-cscs/cscs-docs/pull/39/commits/942d2929e163dc72f4dae62c30d51e38b61ebb8f. I put the note at the top of the page since the public IP is shared in all cases, whether on login nodes or through the proxy server. I also cross-linked a warning in the container engine docs. What do you think?

RMeli: I think that's great, thanks!

docs/guides/internet-access.md
```diff
@@ -0,0 +1,59 @@
...
+    [Authenticating with Docker Hub][ref-ce-third-party-private-registries] makes their rate limit apply per user instead.
+
+## Using SSH through the proxy server 
+
+While use of the proxy server is transparent for most use cases, others need additional configuration for compute nodes.
+An example is cloning git repositories from GitHub over SSH.
+Cloning over https works without additional configuration.
+To make SSH use the proxy server, add the following to your `~/.ssh/config` file:
+
+```bash
```
bcumming, line 33: This isn't bash code, maybe use `terminal`?

You can also give it a title

```
\```terminal "title"
Match Host *,!148.187.0.0/16,!192.168.0.0/16,!172.16.0.0/12,!10.0.0.0/8 exec "hostname -I | grep -vqF 148.187."
    ProxyCommand nc -X connect -x proxy.cscs.ch:8080 %h %p
\```
```

docs/guides/internet-access.md
```diff
@@ -0,0 +1,59 @@
...
+# Internet Access on Alps
+
+The [Alps network][ref-alps-hsn] is mostly configured with private IP addresses (`172.28.0.0/16`).
+Login nodes have public IP addresses which means that they can directly access the internet, while a proxy server provides internet access for compute nodes.
+
+??? info "Compute node proxy configuration"
+
+    Compute nodes are configured with the following environment variables to use the proxy server:
+    
+    ```bash
```
bcumming, line 11: To make it easier for users to cut and paste this when they need to, add `export` to the start of each line.

docs/guides/internet-access.md
bcumming: I prefer the "on Alps"

docs/guides/internet-access.md
```diff
@@ -0,0 +1,59 @@
...
+To make SSH use the proxy server, add the following to your `~/.ssh/config` file:
+
+```bash
+Match Host *,!148.187.0.0/16,!192.168.0.0/16,!172.16.0.0/12,!10.0.0.0/8 exec "hostname -I | grep -vqF 148.187."
+    ProxyCommand nc -X connect -x proxy.cscs.ch:8080 %h %p
+```
+
+This configuration takes into account that login and compute nodes require a different setup.
+
+??? info "Error message when cloning without the proxy set up for SSH"
```
bcumming, line 40: `warning` instead of `info` will make it stand out more?

bcumming (changes requested):

docs/guides/internet-access.md
msimberg: Good catch, indeed. I removed the type altogether, since it's a plain text config file, and added the filename as title in https://github.com/eth-cscs/cscs-docs/pull/39/commits/1f9a772dbbe14a70ffa03e39c65fc2ed45f1b852.

I also found this page which has way more lexers than I would've expected: https://pygments.org/docs/lexers/#. Unfortunately none for ssh config files, but I think plain text will do just fine here.

docs/guides/internet-access.md
msimberg: On one hand I agree, adding the export makes it easier. On the other, do they need to cut and paste if it's configured correctly? This is really just a "FYI". Which makes me think: do we need to keep the actual values? Or just say that they're set appropriately?

docs/guides/internet-access.md
msimberg: My thinking was that it's just "info" as something to show up in search results. The whole section is a "warning" in a sense, saying "warning: to be able to clone on compute nodes you have to set up this config". That said, I'm happy with "warning" as well, really not a big deal for me.

docs/guides/internet-access.md
bcumming: I made this comment because I got burnt by exactly this the other day when using a container - the variables don't seem to be set properly inside the container in some cases.
I wanted to cut and paste from the original docs, and was annoyed by having to manually add `export` for each line.

You could create a table with the variables, and an expandable `example` box that has the text ready to "cut and paste"?

docs/guides/internet-access.md
bcumming: You are right that it isn't a big deal.

I thought that because the box contains an description of an error and what to do, having the orange title with an exclamation mark would make it stand out more on the page.

docs/guides/internet-access.md
msimberg: Fair enough, that's a good reason to leave them here then and add the `export`! I added the `export` in https://github.com/eth-cscs/cscs-docs/pull/39/commits/e5370975c086e777ae159a3b39538ca672d7184a. Is it worth having the extra table as well? I'd prefer not duplicating the information, even if they're close to each other and "easy" to spot if both need to be updated.

docs/guides/internet-access.md
msimberg: That's fair! Made it a warning in https://github.com/eth-cscs/cscs-docs/pull/39/commits/fd29d208357a6e539cfc06e1df9591dcd0b2b6c6. Didn't mean to be difficult.

bcumming (approved):

PR #43, opened 2025-03-14
Better document `CUDA_CACHE_PATH` in CP2K docs

msimberg:
It's set by default on daint (to a different value).

Only reason to keep it here is if we expect many users to be running CP2K on other clusters, but then I'd rather make sure `CUDA_CACHE_PATH` is set by default there instead?

### DIFF
--- a/docs/software/communication/cray-mpich.md
+++ b/docs/software/communication/cray-mpich.md
@@ -7,6 +7,7 @@ It is available through uenvs like [prgenv-gnu][ref-uenv-prgenv-gnu] and [the ap
 The [Cray MPICH documentation](https://cpe.ext.hpe.com/docs/latest/mpt/mpich/index.html) contains detailed information about Cray MPICH.
 On this page we outline the most common workflows and issues that you may encounter on Alps.
 
+[](){#ref-communication-cray-mpich-gpu-aware}
 ## GPU-aware MPI
 
 We recommend using GPU-aware MPI whenever possible, as it almost always provides a significant performance improvement compared to communication through CPU memory.
--- a/docs/software/sciapps/cp2k.md
+++ b/docs/software/sciapps/cp2k.md
@@ -67,6 +67,8 @@ MPS] daemon so that multiple MPI ranks can use the same GPU.
 #SBATCH --uenv=<CP2K_UENV>
 #SBATCH --view=cp2k
 
+export CUDA_CACHE_PATH="/dev/shm/$USER/cuda_cache" # (5)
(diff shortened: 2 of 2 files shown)


bcumming:
I will leave it to @RMeli to comment definitively.

I think that documenting this variable is important - even if it is defined by default on the system.

RMeli:
As I expressed elsewhere, I think it would be beneficial to have a central (and by-platform) documentation of all these variables that are automatically set by us, so that users are aware.

If we want to be consistent with `MPICH_GPU_SUPPORT_ENABLED` not being there, we should probably remove this one too.

However, an alternative that is probably nicer for the end-user would be to explicitly set both (doesn't hurt), and add a note that they are here only for documentation purposes and they are already automatically set in the system? In such case I would do what @msimberg suggest and set it to the actual default `/dev/shm/$USER/cuda_cache/`.

RMeli:
Sorry @msimberg, I misread your suggestion. I agree `CUDA_CACHE` should be set on other clusters too, but documenting the default here doesn't hurt IMO (and if the default changes, it is not a big issue to set it to `/dev/shm/$USER/cuda_cache/`, it will do the job).

abussy:
I agree with @RMeli, leaving it in the documentation does not hurt, even if there is a reasonable default on each cluster. A lot of time would have been saved on the validation of Alps last summer if this kind of stuff had remained in the docs (IIRC, something similar happened when Daint XC was tested, but the knowledge was lost).

msimberg:
All your comments are good! I just pushed a change to restore `CUDA_CACHE_PATH` to the real value set on daint, and to set `MPICH_GPU_SUPPORT_ENABLED` explicitly. I added notes in the annotations.

Do you have a preference for keeping the other box mentioning `MPICH_GPU_SUPPORT_ENABLED`? Or adding another explicit box mentioning `CUDA_CACHE_PATH`?

I added a link to the new "GPU-aware MPI" section from the CP2K info box (https://github.com/eth-cscs/cscs-docs/pull/43/commits/35f48dbe530165a70fa5abdf68254f84da3771bf).

RMeli:
> Do you have a preference for keeping the other box mentioning MPICH_GPU_SUPPORT_ENABLED? Or adding another explicit box mentioning CUDA_CACHE_PATH?

At this point I'd go for consistency and add another mentioning `CUDA_CACHE_PATH`, if it's not too much trouble. Thanks for doing this!

RMeli (approved):
LGTM, thanks!

PR #47, opened 2025-03-17
Update storage section to explain relocatable python venvs with uv

egparedes:


egparedes:
I'm sorry for requesting a review before it was ready (I forgot to commit some changes). Now it should be ready.

RMeli:
> I'm sorry for requesting a review before it was ready

No worries! Review requests have been added automatically because we added ourselves to the `CODEOWNERS` files in #45.

docs/guides/storage.md
```diff
@@ -67,6 +66,28 @@ pip install torch torchvision torchaudio \
 
     In our "simple" pytorch example, I counted **22806 inodes**!
 
+##### Alternative virtual environment creation using uv
```
bcumming, line 69: Instead of adding the `uv` method as a sub-heading under the `pip` method, how about putting them at the same level as tabs?

With one tab for `pip`, and one for `uv` (I would be fine with making the "first" default tab the `uv` one, so that we can push users towards `uv`.

https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#usage

bcumming (changes requested):

docs/guides/storage.md
bcumming: See here for an example of what tabs look like in our docs
https://eth-cscs.github.io/cscs-docs/software/tools/linaro/#quickstart-guide

docs/guides/storage.md
```diff
@@ -105,27 +144,39 @@ mksquashfs $SCRATCH/sqfs-demo/.pyenv pyenv.squashfs \
...
-Note that the original virtual environment is still installed in `$SCRATCH/sqfs-demo/.pyenv`, however the squashfs image has been mounted on top of it, so the single squashfs file is being accessed instead of the many files in the original version.
+    ```bash
+    cd $SCRATCH/sqfs-demo
+    uenv start --view=default \
+        prgenv-gnu/24.11:v1,$PWD/py_venv.squashfs:$SCRATCH/sqfs-demo/.venv
+    source .venv/bin/activate
+    ```
 
-A benefit of this approach is that the squashfs file can be copied to a location that is not subject to the Scratch cleaning policy.
+    Remember that virtual environments created by `uv` are relocatable only if the `--relocatable` option flag is passed to the `uv venv` command as mentioned in step 1. In that case, the generated environment is relocatable and thus it is possible to mount it in multiple locations without problems.
```
RMeli, line 156:
```suggestion
    Remember that virtual environments created by `uv` are relocatable only if the `--relocatable` flag is passed to the `uv venv` command as mentioned in step 1. In that case, the generated environment is relocatable and thus it is possible to mount it in multiple locations without problems.
```

docs/guides/storage.md
```diff
@@ -105,27 +144,39 @@ mksquashfs $SCRATCH/sqfs-demo/.pyenv pyenv.squashfs \
...
+    ```bash
+    cd $SCRATCH/sqfs-demo
+    uenv start --view=default \
+        prgenv-gnu/24.11:v1,$PWD/py_venv.squashfs:$SCRATCH/sqfs-demo/.venv
+    source .venv/bin/activate
+    ```
+
+    Note that the original virtual environment is still installed in `$SCRATCH/sqfs-demo/.venv`, however the squashfs image has been mounted on top of it, so the single squashfs file is being accessed instead of the many files in the original version.
+
+    A benefit of this approach is that the squashfs file can be copied to a location that is not subject to the Scratch cleaning policy.
```
RMeli, line 169:
```suggestion
    A benefit of this approach is that the squashfs file can be copied to a location that is not subject to the scratch cleaning policy.
```
or
```suggestion
    A benefit of this approach is that the squashfs file can be copied to a location that is not subject to the `$SCRATCH` cleaning policy.
```

bcumming (approved):
Fantastic, thanks!

PR #53, opened 2025-03-21
Add basic style guide

msimberg:
This adds a basic `STYLE.md` in the root of the repo as a basic style guide (the filename and location is not a convention as far as I know, I just thought it fits). I've tried to capture most of the current style. That doesn't mean we have to keep it this way. Bikeshed away with review comments!

It'd be nice to have this in the actual docs as well, but I wasn't sure where to put it yet so it's currently outside the main `docs` tree.


STYLE.md
```diff
@@ -0,0 +1,48 @@
...
+They should be followed for most cases, but as a guideline it can be broken, _with good reason_.
+
+The documentation is written using [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/).
+
+## General Formatting
+
+- Format paragraphs with one sentence per line for easier diffs.
+- Leave a space before and after headings.
+
+## Headings are Written in Title Case
```
msimberg, line 13: This seems to be the case for most headings, but is not completely consistent. My purely subjective opinion is sentence case for headings as well, but don't care too much as long as we stick to one style.

STYLE.md
```diff
@@ -0,0 +1,48 @@
...
+Nesting headings up to three levels is generally ok.
+
+## Unordered and Ordered Lists
+
+Write lists as proper sentences.
+Separate the items simply with commas if each item is simple, or make each item a full sentence if the items are longer and contain multiple sentences.
+
+1. The first item can look like this,
+2. The second like this, and
+3. The third item like this.
```
msimberg, line 28:
```suggestion
2. the second like this, and
3. the third item like this.
```
?

STYLE.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Using Admonitions
+
+Aim to include examples, notes, warnings using [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) whenever appropriate.
+They stand out better from the main text, and can be collapsed by default if needed.
+
+!!! example "Example one"
```
msimberg, line 35: Some admonition titles use all lower case, some sentence case. Preference?

RMeli:
> It'd be nice to have this in the actual docs as well, but I wasn't sure where to put it yet so it's currently outside the main docs tree.

I'll review later, but we already have a "Contributing" section, therefore it could go there, or as a subsection of that?

STYLE.md
RMeli: I've been doing and suggesting the opposite... But I also do not care as long as we are consistent.

docs/contributing/index.md
```diff
@@ -92,3 +92,50 @@ Once your changes are ready, click on the "Commit changes..." button in the top
 
     * if the change is small and you are CSCS staff, you can merge the PR immediately
     * all other changes can be
+
+## Style Guide
+
+This section contains general guidelines for how to format and present documentation in this repository.
+They should be followed for most cases, but as a guideline it can be broken, _with good reason_.
```
msimberg, line 99: I'm not sure if this is too inviting to breaking the guidelines.

msimberg:
I moved the style guide to a subsection in `contributing.md`. Could also be a sub-page. Let me know what you prefer.

bcumming:
Should the style guide be integrated into the "contributing" section of the docs?

I say yes!

bcumming:
This overlaps with another PR made at about the same time, that discusses "how to link" and provides justification for why we don't break sentences.

#54

docs/contributing/index.md
```diff
@@ -92,3 +92,50 @@ Once your changes are ready, click on the "Commit changes..." button in the top
...
+Nesting headings up to three levels is generally ok.
+
+### Unordered and Ordered Lists
+
+Write lists as proper sentences.
+Separate the items simply with commas if each item is simple, or make each item a full sentence if the items are longer and contain multiple sentences.
+
+1. The first item can look like this,
+2. The second like this, and
+3. The third item like this.
```
bcumming, line 121: I would put the `and` at the start of the third line, and star each line with a lower case letter (the whole list is a sentence, no?)

I agree that care should be taken with formating lists, but it is tricky to canonicalize without being too strict. For example, I would format your list as follows:

contain multiple sentences:

1. the first item can look like this;
2. the second like this;
3. and the third like this.

msimberg: > but it is tricky to canonicalize without being too strict
This is probably key, so in the end these are just guidelines.

Agree on capitalization (see https://github.com/eth-cscs/cscs-docs/pull/53#discussion_r2007514552). I feel like it's more common to have the "and" or "or" trailing on the second-to-last item though?

bcumming: Honestly, I don't know.
It feels wrong to me, because if we use the oxford comma (as we should), the and comes after the comma.

I think if people go to enough effort to punctuate their lists with care, that's enough?

STYLE.md
bcumming: Personally I am more comfortable with lower case if I had to choose (except for the first word in the heading).

Of course, I am almost certainly wildly inconsistent in practice.

docs/contributing/index.md
```diff
@@ -192,3 +192,45 @@ Once your changes are ready, click on the "Commit changes..." button in the top
...
+1. The first item can look like this,
+2. the second like this, and
+3. the third item like this.
+
+### Using Admonitions
+
+Aim to include examples, notes, warnings using [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) whenever appropriate.
+They stand out better from the main text, and can be collapsed by default if needed.
+
+!!! example "Example one"
```
RMeli, line 223: Which style should be used for the title?

docs/contributing/index.md
```diff
@@ -192,3 +192,45 @@ Once your changes are ready, click on the "Commit changes..." button in the top
...
+    This is an example.
+    The title of the example uses [sentence case](https://en.wikipedia.org/wiki/Letter_case#Sentence_case).
+    
+??? note "Collapsed note"
+    This note is collapsed, because it uses `???`.
+
+[](){#ref-style-references}
+### References
```
RMeli, line 231: Maybe worth adding a quick note on how to add a reference? Or a link to the documentation?

docs/contributing/index.md
msimberg: The text in the example says sentence case, because it's also my preference here.

docs/contributing/index.md
msimberg: I should remove this section now. @bcumming already added more detailed docs in #54.

msimberg: Removed.

docs/contributing/index.md
msimberg: If it's all the same to you, I'd leave it with the trailing and/or on the second-to-last item, but indeed, let's not _actually_ bikeshed this too much.

> I think if people go to enough effort to punctuate their lists with care, that's enough?

Yes, of course.

STYLE.md
msimberg: Of course not a big deal. This is just aiming to bring a bit of consistency, but we'll break the rules anyway, intentionally or unintentionally.

STYLE.md
msimberg: https://github.com/eth-cscs/cscs-docs/pull/53/commits/dce66c6957cc6d9cbf08bd20eea423c99c824418

docs/contributing/index.md
RMeli: Sorry, I missed it was part of the example.

RMeli (approved):

bcumming (approved):

PR #59, opened 2025-03-26
First commit of OpenMP documentation.

jpcoles-cscs:
These changes specifically address the recommended setting based on santis using Low Noise Mode.


bcumming (changes requested):
This documentation provides advice on how to configure the environment and slurm for running jobs, so I think it should ultimately go into the "running jobs" guide.

However, for our immediate purposes, I would crate a page under "guides" for just the Gordon Bell runs in March-April 2025, that contains all of the information specific to the configuration of Santis, and move those docs to the relevant sections when these changes are rolled out across the whole system.

PR #70, opened 2025-04-01
Clarify uenv spack view

RMeli:
I think the current description of the `spack` view is misleading. I tried to improve things, trying to make clear that the `spack` view sets some environment variables that are _not_ automatically used by Spack.

### DIFF
--- a/docs/build-install/uenv.md
+++ b/docs/build-install/uenv.md
@@ -49,7 +49,7 @@ To use Spack as an upstream, the uenv can be started with the `spack` view:
 uenv start prgenv-gnu/24.11:v1 --view=spack
 ```
 
-??? note "What does the `spack` view do?"
+!!! note "What does the `spack` view do?"
     The `spack` view sets environment variables that provide information about the version of Spack that was used to build the uenv, and where the uenv Spack configuration is stored.
     
     | <div style="width:12em">variable</div> | <div style="width:18em">example</div> | description |
--- a/docs/build-install/uenv.md
+++ b/docs/build-install/uenv.md
@@ -49,7 +49,7 @@ To use Spack as an upstream, the uenv can be started with the `spack` view:
 uenv start prgenv-gnu/24.11:v1 --view=spack
 ```
 
-!!! note "What does the `spack` view do?"
+??? note "What does the `spack` view do?"
     The `spack` view sets environment variables that provide information about the version of Spack that was used to build the uenv, and where the uenv Spack configuration is stored.
     
     | <div style="width:12em">variable</div> | <div style="width:18em">example</div> | description |
--- a/docs/build-install/uenv.md
+++ b/docs/build-install/uenv.md
@@ -65,7 +65,7 @@ uenv start prgenv-gnu/24.11:v1 --view=spack
         Therefore, they don't change Spack-related environment variables.
         You can use them to consistently set Spack-related environment variables.
 
-You can then set the uenv as an upstream Spack instance as follows:
+You can then set the uenv as an [upstream Spack instance](https://spack.readthedocs.io/en/latest/chain.html) as follows:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/build-install/uenv.md
```diff
@@ -43,24 +43,38 @@ graph TD
...
     | `UENV_SPACK_URL`         | `https://github.com/spack/spack.git` | The git repository for Spack - nearly always the main spack/spack repository. |
     | `UENV_SPACK_COMMIT`      | `c6d4037758140fe...0cd1547f388ae51` | The commit of Spack that was used |
 
+    !!! warning
+
+        The environment variables set by the `spack` view are scoped by `UENV_`.
+        Therefore, they don't change Spack-related environment variables.
+        You can use them to consistently set Spack-related environment variables.
+
+You can then set the uenv as an upstream Spack instance as follows:
```
RMeli, line 68:
```suggestion
You can then set the uenv as an [upstream Spack instance](https://spack.readthedocs.io/en/latest/chain.html) as follows:
```

docs/build-install/uenv.md
```diff
@@ -43,24 +43,38 @@ graph TD
...
 
-To use Spack as an upstream, the uenv has to be started with the `spack` view:
+To use Spack as an upstream, the uenv can be started with the `spack` view:
 
 ```bash
 uenv start prgenv-gnu/24.11:v1 --view=spack
 ```
 
-??? note "what does the `spack` view do?"
+??? note "What does the `spack` view do?"
```
RMeli, line 52:
```suggestion
!!! note "What does the `spack` view do?"
```

docs/build-install/uenv.md
```diff
@@ -43,24 +43,38 @@ graph TD
...
 
-To use Spack as an upstream, the uenv has to be started with the `spack` view:
+To use Spack as an upstream, the uenv can be started with the `spack` view:
 
 ```bash
 uenv start prgenv-gnu/24.11:v1 --view=spack
 ```
 
-??? note "what does the `spack` view do?"
+!!! note "What does the `spack` view do?"
```
RMeli, line 52:
```suggestion
??? note "What does the `spack` view do?"
```

bcumming (approved):

docs/build-install/uenv.md
```diff
@@ -43,24 +43,38 @@ graph TD
...
+    !!! warning
+
+        The environment variables set by the `spack` view are scoped by `UENV_`.
+        Therefore, they don't change Spack-related environment variables.
+        You can use them to consistently set Spack-related environment variables.
+
+You can then set the uenv as an [upstream Spack instance](https://spack.readthedocs.io/en/latest/chain.html) as follows:
+
+```bash
+export SPACK_SYSTEM_CONFIG_PATH=$UENV_SPACK_CONFIG_PATH
```
msimberg, line 71: This is good, but may be dangerous if users don't know to use separate indexes, user config paths, etc. per uenv. I don't know if we want/need to provide a full guide on how to do that (I can share what I do), but a warning may be useful at least at this point?

RMeli: Merged too quickly. =)

I actually did not realize that the reminder of the documentation was about `env-spack`; I thought it was still very much WIP and undocumented. Therefore, the above is probably bad for the end user, which should be encouraged to use the tool (I suppose). I can rephrase it to make it clearer that things are not set automatically but one should do so if they wish (without giving too many suggestions).

msimberg: I didn't realize either :grimacing: Yeah, I think the idea is that the `uenv-spack` tool consumes `UENV_SPACK_CONFIG_PATH` etc. but it sets up a clone of spack etc. on its own, right? If that's the case, exporting `SPACK_SYSTEM_CONFIG_PATH` here probably isn't going to harm anyone, but it won't help either...

PR #71, opened 2025-04-01
Improve uenv-spack documentation

RMeli:
Follow-up to #70.

### DIFF
--- a/docs/build-install/uenv.md
+++ b/docs/build-install/uenv.md
@@ -1,6 +1,8 @@
-Uenv are user environments that provide scientific applications, libraries and tools on Alps. This article use them to build software.
+Uenv are user environments that provide scientific applications, libraries and tools on [Alps][ref-alps]. 
+This article explains how to use them to build software.
 
-For more documentation on how to find, download and use uenv in your workflow, see the [env tool documentation][ref-uenv].
+For more documentation on how to find, download and use uenv in your workflow,
+see the [uenv documentation][ref-uenv].
 
 [](){#ref-building-uenv-spack}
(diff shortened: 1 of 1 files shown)


docs/build-install/uenv.md
```diff
@@ -65,35 +70,46 @@ uenv start prgenv-gnu/24.11:v1 --view=spack
...
-export SPACK_SYSTEM_CONFIG_PATH=$UENV_SPACK_CONFIG_PATH
-```
-
 ??? warning "Upstream Spack version"
 
     It is strongly recomended that your version of Spack and the version of Spack in the uenv match when building software on top of an uenv.
 
+!!! note "Advanced Spack users"
+
+    Advanced Spack users can use the environment variables set by the `spack` view to manually configure the uenv as a Spack upstream instance.
```
msimberg, line 79: Should we at least add:
```suggestion
    Advanced Spack users can use the environment variables set by the `spack` view to manually configure the uenv as a Spack upstream instance.
    If using multiple uenvs, we recommend using a different spack instance per uenv.
```
?

msimberg (approved):

docs/build-install/uenv.md
RMeli:
```suggestion
    Advanced Spack users can use the environment variables set by the `spack` view to manually configure the uenv as a Spack upstream instance.
    
    !!! tip
        If using multiple uenvs, we recommend using a different Spack instance per uenv.
```

msimberg (approved):
Thanks @RMeli!

PR #57, opened 2025-03-24
Fix typos

msimberg:
Fixes the most obvious misspellings.

PR #74, opened 2025-04-03
Update code blocks and add guidelines on usage to style guide

msimberg:
I've attempted to fix up some code blocks. Whatever was obviously a terminal (ehm, console) session I've updated to use `console`. Turns out `terminal` isn't a valid lexer, and just does no highlighting. I've changed some types to `bash` or `console` a bit depending on the contents of the code block, but I have not tried to fix everything.

I've added a section about code blocks to the style guide, which we can try to follow when updating existing or adding new docs.

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -13,20 +13,20 @@ We use the GitHub fork and pull request model for development:
 Clone your fork repository on your PC/laptop:
 ```bash
 # clone your fork of the repository
-> git clone git@github.com:${githubusername}/cscs-docs.git
-> cd cscs-docs
-> git switch -c 'fix/ssh-alias'
+git clone git@github.com:${githubusername}/cscs-docs.git
+cd cscs-docs
+git switch -c 'fix/ssh-alias'
 # ... make your edits ...
 # add and commit your changes
-> git add <files>
-> git commit -m 'update the ssh docs with aliases for all user lab vclusters'
-> git push origin 'fix/ssh-alias'
+git add <files>
+git commit -m 'update the ssh docs with aliases for all user lab vclusters'
+git push origin 'fix/ssh-alias'
 ```
 Then navigate to GitHub, and create a pull request.
 
 The `serve` script in the root path of the repository can be used to view the docs locally:`
-```
-> ./serve
+```bash
+./serve
 ...
 INFO    -  [08:33:34] Serving on http://127.0.0.1:8000/
 ```
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -236,38 +236,48 @@ The documentation uses [pygments](https://pygments.org) for highlighting.
 See [list of available lexers](https://pygments.org/docs/lexers/#) for the languages that you can use for code blocks.
 
 Use [`console`](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer) for interactive sessions with prompt-output pairs:
-````markdown
-```console title="Hello, world!"
-$ echo "Hello, world!"
-Hello, world!
-```
-````
 
-The above becomes:
+=== "Markdown"
 
-```console title="Hello, world!"
-$ echo "Hello, world!"
-Hello, world!
-```
+    ````markdown
+    ```console title="Hello, world!"
+    $ echo "Hello, world!"
+    Hello, world!
+    ```
+    ````
+
+=== "Rendered"
+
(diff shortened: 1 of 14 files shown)


docs/contributing/index.md
```diff
@@ -228,3 +228,46 @@ They stand out better from the main text, and can be collapsed by default if nee
...
 If an admonition is collapsed by default, it should have a title.
+
+### Code blocks
+
+Use [code blocks](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/) when you want to display monospace text in a programming language, terminal output, configuration files etc.
+The documentation uses [pygments](https://pygments.org) for highlighting.
+See [list of available lexers](https://pygments.org/docs/lexers/#) for the languages that you can use for code blocks.
+
+Use [`console`](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer) for interactive sessions with prompt-output pairs:
+````markdown
```
bcumming, line 239: You could use tabs for this?

docs/contributing/index.md
```diff
@@ -228,3 +228,46 @@ They stand out better from the main text, and can be collapsed by default if nee
...
+```bash title="Hello, world!"
+echo "Hello, world!"
+```
+````
+
+The above becomes:
+
+```bash title="Hello, world!"
+echo "Hello, world!"
+```
```
bcumming, line 273: From these docs and updates on other pages, it looks like you lean towards using `$` for the prompt in shell code blocks instead of `$`.

How about making this a formal policy, and documenting it here?

docs/guides/gb2025.md
```diff
@@ -45,9 +45,9 @@ The consequence of this setting is that only 71 cores per socket can be requeste
 !!! warning "Unable to allocate resources: Requested node configuration is not availabl"
```
bcumming, line 45:
```suggestion
!!! warning "Unable to allocate resources: Requested node configuration is not available"
```

docs/software/container-engine.md
```diff
@@ -50,13 +50,13 @@ Since the `ubuntu.toml` file is located in the [EDF search path][ref-ce-edf-sear
     The above terminal snippet demonstrates how to launch a containerized environment using Slurm with the `--environment` option.
     Click on the :fontawesome-solid-circle-plus: icon for information on each command.
 
-    ```bash
-    daint-ln002 > srun --environment=ubuntu --pty bash   # (1)
+    ```console
+    [daint-ln002]$ srun --environment=ubuntu --pty bash   # (1)
```
bcumming, line 54: The original docs in KB had very long prompts, so when the docs were ported I tried to shorten all of them to a single `>`.

In this case adding the hostname to the prompt is probably useful, because it provides additional context about whether the shell is on the login node / compute node.

bcumming (changes requested):
Great!
My only suggested improvement is to document the use of `$` for "user prompt" in console/bash examples.

docs/contributing/index.md
msimberg: Yeah, sorry, it was actually not so much a preference, but a must. `>` doesn't trigger correct highlighting, whereas pygments recognizes `$` as the actual prompt.

I'll make it clear that `$` is actually a requirement, not just a preference.

docs/guides/gb2025.md
msimberg: Nice, one of many typos that I've missed in #57.

docs/software/container-engine.md
msimberg: Yeah, I've tried not to modify the code blocks significantly, but here the prompt indeed makes sense.

The only thing I didn't mention: it seems like `daint-ln002$ ...` does _not_ trigger highlighting either, but (e.g.) `[daint-ln002]$` does, which is why I did more than just change `>` to `$`. I don't know what the exact rules are unfortunately, and I'm leaving that for future investigation.

docs/contributing/index.md
msimberg: https://github.com/eth-cscs/cscs-docs/pull/74/commits/12475300572e16f29aac53be0a9653cf3ab652a5

docs/contributing/index.md
msimberg: https://github.com/eth-cscs/cscs-docs/pull/74/commits/12475300572e16f29aac53be0a9653cf3ab652a5

RMeli:
Thanks @msimberg!

PR #92, opened 2025-04-15
remove gordon bell 2025 docs

bcumming:


msimberg (changes requested):
Looks good, but the santis page still links to the GB guide (and the build luckily fails).

msimberg (approved):

PR #99, opened 2025-04-23
document CPE in a container

finkandreas:


bcumming (approved):
Thanks Andreas.

I will go ahead and merge these, then do some refactoring of where they fit into the docs.

PR #106, opened 2025-05-05
Update pytorch.md

vondele:
It is probably good to explicitly add `--system-site-packages` to the venv creation step


vondele:
I think one notification of the preview would have been enough `@bot`

lukasgd:
I was actually asking for this in https://github.com/eth-cscs/cscs-docs/pull/84#discussion_r2042661895, but then that PR got merged without finishing the discussions. Maybe we can do this here?

msimberg (approved):

PR #107, opened 2025-05-05
Add jupyterlab docs

rsarm:


msimberg (changes requested):
This is very nice, thanks @rsarm!

It looks like most of the content between the daint and eiger pages is identical or at least similar. Do you think it'd make sense to combine the two into one page (probably directly on the root jupyterlab page?), only highlighting the differences? You could use tabs like e.g. https://eth-cscs.github.io/cscs-docs/software/devtools/linaro-uenv/#install-and-configure-the-linaro-client-on-your-local-machine (source: https://github.com/eth-cscs/cscs-docs/blob/77231248428c85f966686ed5e5b8e130f0875617/docs/software/devtools/linaro-uenv.md?plain=1#L102-L159).

Also, would you mind adding yourself to `CODEOWNERS` for these pages?

bcumming (changes requested):
Thanks for the contribution!

A general comment is to check out the contributing guide:
    https://eth-cscs.github.io/cscs-docs/contributing/

For example: https://eth-cscs.github.io/cscs-docs/contributing/#text-formatting

Also, because these docs will replace the KB docs when Eiger has been redeployed in the same style as Daint, we don't need to include docs for "old Eiger"

msimberg:
> I removed the docs for Eiger. I had the two separate versions because Eiger's deployment is going to be changed soon.

:+1:

> I couldn't find how to set the `CODEOWNERS`. Could you point to an example?

Add your github username to this file: https://github.com/eth-cscs/cscs-docs/blob/main/.github/CODEOWNERS.

> Regarding code blocks: I didn't put the `$` prompt on the shell commands to make simpler copying and pasting it to a terminal. Also, for single-line commands such as `ln -s $SCRATCH $HOME/scratch` or `pip install ipcmagic-cscs` I didn't put a title, since they are clear. Is that ok? I don't mind adding the titles and the prompt. Just wanted to ask first.

I think what you've done follows the guidelines just fine for exactly the reasons you described. Rule 0 is: they're just guidelines.

docs/services/jupyterlab.md
```diff
@@ -0,0 +1,104 @@
+[](){#ref-jlab}
+# JupyterLab
+
+## Access and setup
+
+The JupyterHub service enables the interactive execution of JupyterLab on [Daint][ref-cluster-daint] on a single compute node.
+
```
msimberg, line 7: Not a must, just thinking out loud: would it still make sense to add a `!!! note` saying something about jypyterhub on eiger?

.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 * @bcumming @msimberg @RMeli
+docs/services/jupyterlab @rsarm
```
msimberg, line 2:
```suggestion
docs/services/jupyterlab.md @rsarm
```

rsarm:
From my part this is ready.

msimberg (approved):

bcumming (approved):

PR #109, opened 2025-05-09
add warning that the cicd docs are incomplete

finkandreas:
Since some people are sharing the link to the new docs for CI/CD, I've added a warning at the top, that they have not been fully ported.


finkandreas:
Yeah, the blanket warning on a black background fades away in my normal view. With the current status (and given that the CI/CD docs look "long"), one could think that CI/CD is already fully ported. I just want to make sure, people are aware that there's still more docs at the old space.

msimberg (approved):

PR #126, opened 2025-06-02
add docs on building GROMACS from source

kanduri:


RMeli (approved):
Thanks @kanduri, LGTM

PR #145, opened 2025-06-06
MLP storage advice

nickjbrowning:


bcumming (changes requested):
This page provides information about file systems that is generic to all.

This information is MLP specific (or ML workflow specific), so it doesn't quite fit in her.

I would suggest creating an "ML Storage" Guide under the Guides section?

nickjbrowning:
> This page provides information about file systems that is generic to all.
> 
> This information is MLP specific (or ML workflow specific), so it doesn't quite fit in her.
> 
> I would suggest creating an "ML Storage" Guide under the Guides section?
> 
> you can then link to it from the storage docs for MLP: https://docs.cscs.ch/platforms/mlp/#file-systems-and-storage

Why does this not fit? This is MLP-specific advice on the MLP page.

bcumming:
The proposed updates directly discuss best practices for where to store checkpoints and inputs for LLM training. Specifically it also discusses Iopsstore, which is currently only for the use of MLP - suggesting that all users on the system should start using it would affect LLM training runs.

More generally, the documentation on this page aims to explain what Scratch/Store/Home are, and the policies that apply to them.

We need additional best-practices guides for storage. For the MLP you have practices, and I think these can go in a page under the MLp, where they are easily acccessible to MLP users.

A more generic guide, that covers topics like LUSTRE striping, can go under the Guides section.

bcumming (approved):
Sorry, I completely missed that these updates were made to the MLP storage docs.

Please ignore my earlier comments

PR #146, opened 2025-06-06
Add more environment variables for NCCL

msimberg:
Draft. Not clear if all are needed.

Do we need to start separating recommended environment variables by nccl, libfabric, etc. version, or is it sufficient if we recommend the best practices for the latest versions?

### DIFF
--- a/docs/software/communication/nccl.md
+++ b/docs/software/communication/nccl.md
@@ -18,11 +18,20 @@ While the container engine sets these automatically when using the NCCL hook, th
 
 ```bash
 export NCCL_NET="AWS Libfabric" # (1)!
+export NCCL_CROSS_NIC="1"
+export NCCL_NET_FORCE_FLUSH="1"
 export NCCL_NET_GDR_LEVEL=PHB # (2)!
 export FI_CXI_DEFAULT_CQ_SIZE=131072 # (3)!
 export FI_CXI_DEFAULT_TX_SIZE=32768
 export FI_CXI_DISABLE_HOST_REGISTER=1
+export FI_CXI_RDZV_EAGER_SIZE="0"
+export FI_CXI_RDZV_GET_MIN="0"
+export FI_CXI_RDZV_THRESHOLD="0"
+export FI_CXI_RX_MATCH_MODE="hardware"
 export FI_CXI_RX_MATCH_MODE=software
+export FI_CXI_SAFE_DEVMEM_COPY_THRESHOLD="16777216"
+export FI_MR_CACHE_MAX_COUNT="524288"
+export FI_MR_CACHE_MAX_SIZE="-1"
 export FI_MR_CACHE_MONITOR=userfaultfd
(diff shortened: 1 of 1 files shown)


docs/software/communication/nccl.md
```diff
@@ -18,11 +18,20 @@ While the container engine sets these automatically when using the NCCL hook, th
 
 ```bash
 export NCCL_NET="AWS Libfabric" # (1)!
+export NCCL_CROSS_NIC=1
+export NCCL_NET_FORCE_FLUSH=1
 export NCCL_NET_GDR_LEVEL=PHB # (2)!
+export NCCL_SOCKET_IFNAME=hsn
```
msimberg, line 24: NCCL_CROSS_NIC and NCCL_SOCKET_IFNAME are set in the CE hook, and seem safe to recommend for most users (latter anyway seems to be just a sanity setting to avoid using the wrong network).

boeschf:
we should also change https://github.com/msimberg/cscs-docs/blob/63149793755baedc9052b2e4aad920f01b266f33/docs/software/ml/pytorch.md?plain=1#L318

msimberg:
> we should also change https://github.com/msimberg/cscs-docs/blob/63149793755baedc9052b2e4aad920f01b266f33/docs/software/ml/pytorch.md?plain=1#L318

Good catch. I wonder if we can avoid copy-pasting and manually having to make sure they're synchronized. I like that the pytorch submission script is standalone. Do you think it would be bad if we just link to the nccl page from there? I'd imagine many users will miss copying the nccl variables in that case...

From a quick search this also seems to exist: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files. That might allow defining these in one place and including in many.

That said, that might be overkill at the moment so I might just copy them over for now.

Any comments on which vars we can actually safely recommend and which we might want to wait with still? Or if we have to add a warning about some variables only being good/useful with nccl 2.26 and libfabric 1.22?

docs/software/communication/nccl_env_vars
```diff
@@ -0,0 +1,27 @@
...
+# TCP, which would be significantly slower than with the plugin. More information
+# about `NCCL_NET` can be found at
+# https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-net.
+export NCCL_NET="AWS Libfabric"
+# Use GPU Direct RDMA when GPU and NIC are on the same NUMA node. More
+# information about `NCCL_NET_GDR_LEVEL` can be found at
+# https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-net-gdr-level-formerly-nccl-ib-gdr-level).
+export NCCL_NET_GDR_LEVEL=PHB
+export NCCL_CROSS_NIC=1
+export NCCL_NET_FORCE_FLUSH="1"
```
msimberg, line 13:
```suggestion
export NCCL_NET_FORCE_FLUSH=1
```

PR #148, opened 2025-06-11
Slurm over-subscription

bcumming:
Based on internal Confluence notes: https://confluence.cscs.ch/spaces/VCUE/pages/894961135/Parallel+srun+calls

I have put this in a branch on the main repo for @jpcoles-cscs to easily contribute to.


finkandreas:
I would add in the example to the srun calls also the flag `--output "out-%J.log` for a reasonable default for stdout/stderr redirection of each srun call.

I think it's easier for people to remove flags, instead of searching for the correct flag in the manpages (although `--output` and `--error` should be quite well known).

jpcoles (approved):

PR #150, opened 2025-06-11
Reenable spell checker

msimberg:
Still testing...


msimberg:
I'm going to force merge this for testing. With the `pull_request_target` event this workflow isn't going to run until it's merged to `main`. The workflow runs somewhat ok with `pull_request` but can't comment on the PR. Merging to make sure that the `pull_request_target` event also works.

PR #153, opened 2025-06-16
improve VS Code docs when using more than one cluster

bcumming:
Rendered docs:
https://docs.tds.cscs.ch/153/access/vscode/


kanduri (approved):
LGTM! The doc looks very high quality now, not like a Confluence/Google Docs shared doc.

PR #166, opened 2025-06-24
Add lustre tuning guide

fawzi:
Added the basic info and commands to improve lustre performance.
I did not describe things that we do not use (like pool), or that normally should not be used like index/offset.

### DIFF
--- a/docs/guides/storage.md
+++ b/docs/guides/storage.md
@@ -114,25 +114,25 @@ To set up a default so all newly created folders and dirs inside or your desired
     For more information read the setfacl man page: `man setfacl`.
 
 [](){#ref-guides-storage-lustre}
-## Lustre Tuning
-[Capstor][ref-alps-capstor] and [Iopsstor][ref-alps-iopsstor] are both [lustre](https://lustre.org) filesystem.
-
-As shown in the schema below
+## Lustre tuning
+[Capstor][ref-alps-capstor] and [Iopsstor][ref-alps-iopsstor] are both [Lustre](https://lustre.org) filesystem.
 
 ![Lustre architecture](../images/storage/lustre.png)
 
-Lustre uses *metadata* servers to store and query metadata which is basically what is shown by `ls`: directory structure, file permission, modification dates,...
+As shown in the schema above, Lustre uses *metadata* servers to store and query metadata, which is basically what is shown by `ls`: directory structure, file permission, and modification dates.
 Its performance is roughly the same on [Capstor][ref-alps-capstor] and [Iopsstor][ref-alps-iopsstor].
-This data is globally synchronized, which means that handling many small files is not especially suited for Lustre, see the discussion on [how to handle many small files][ref-guides-storage-small-files].
+This data is globally synchronized, which means Lustre is not well suited to handling many small files, see the discussion on [how to handle many small files][ref-guides-storage-small-files].
 
 The data itself is subdivided in blocks of size `<blocksize>` and is stored by Object Storage Servers (OSS) in one or more Object Storage Targets (OST).
-The blocksize and number of OSTs to use is defined by the striping settings.
-A new file or directory ihnerits them from its parent directory.
-The `lfs getstripe <path>` command can be used to get information on the actual stripe settings.
-For directories and empty files `lfs setstripe --stripe-count <count> --stripe-size <size> <directory/file>` can be used to set the layout. The simplest way to have the correct layout is to copy to a directory with the correct layout
+The blocksize and number of OSTs to use is defined by the striping settings, which are applied to a path, with new files and directories ihneriting them from their parent directory.
+The `lfs getstripe <path>` command can be used to get information on the stripe settings of a path.
+For directories and empty files `lfs setstripe --stripe-count <count> --stripe-size <size> <directory/file>` can be used to set the layout.
+The simplest way to have the correct layout is to copy to a directory with the correct layout
 
-A blocksize of 4MB gives good throughput, without being overly big, so it is a good choice when reading a file sequentially or in large chunks, but if one reads shorter chunks in random order it might be better to reduce the size, the performance will be smaller, but the performance of your application might actually increase.
-https://doc.lustre.org/lustre_manual.xhtml#managingstripingfreespace
+!!! tip "A blocksize of 4MB gives good throughput, without being overly big..."
+    ... so it is a good choice when reading a file sequentially or in large chunks, but if one reads shorter chunks in random order it might be better to reduce the size, the performance will be smaller, but the performance of your application might actually increase.
+    See the [Lustre documentation](https://doc.lustre.org/lustre_manual.xhtml#managingstripingfreespace) for more information.
+
 
 !!! example "Settings for large files"
     ```console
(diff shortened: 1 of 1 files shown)


fawzi:
probably we should link to it from MLP and maybe other places too

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
+# Lustre Tuning
+`/capstor/` and `/iopsstor` are both [lustre](https://lustre.org) filesystem.
```
msimberg, line 2:
```suggestion
# Lustre tuning

[Capstor][ref-alps-capstor] and [Iopsstor][ref-alps-iopsstor] are both [Lustre](https://lustre.org) filesystems.
```

In general, do have a quick read of https://docs.cscs.ch/contributing/ for some guidelines on formatting etc.

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+Lustre is an open-source, parallel file system used in HPC systems.
```
msimberg, line 3: I think this sentence could be left out for brevity. It doesn't add much to the specific goal of teaching users about tuning the filesystem IMO.

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+As shown in ![Lustre architecture](/images/storage/lustre.png) uses *metadata* servers to store and query metadata which is basically what is shown by `ls`: directory structure, file permission, modification dates,..
```
msimberg, line 4: This image doesn't show up in the rendered output. Could you make sure that it's correctly rendered?

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+This data is globally synchronized, which means that handling many small files is not especially suited for lustre, and the perfomrance of that part is similar on both capstor and iopsstor.
```
msimberg, line 5:
```suggestion
This data is globally synchronized, which means that handling many small files is not especially suited for Lustre, and the perfomrance of that part is similar on both capstor and iopsstor.
```
here and elsewhere.

msimberg:
```suggestion
This data is globally synchronized, which means that handling many small files is not well suited for lustre.
```
Leave out the second part? Iopsstor should handle small files slightly better than Capstor, no? I may very well also be mistaken in which case it's obviously good to leave it in in.

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+With many small files, a local filesystems like `/dev/shmem/$USER` or "/tmp", if enough memory can be spared for it, can be *much* faster, and offset the packing/unpacking work. Alternatively using a squashed filesystems can be a good option.
```
msimberg, line 6:
```suggestion
With many small files, an in-memory filesystem like `/dev/shm/$USER` or "/tmp", if enough memory can be spared for it, can be *much* faster, and offset the packing/unpacking work. Alternatively using a squashed filesystems can be a good option.
```
?

msimberg: There's a lot of overlap with https://docs.cscs.ch/guides/storage/#many-small-files-vs-hpc-file-systems. Can we link or avoid some duplication of the motivation etc.?

Actually, do you think this guide could fit inside the existing storage guide?

docs/guides/lustre-tuning.md
msimberg: Some of this motivation overlaps with https://docs.cscs.ch/guides/storage/#many-small-files-vs-hpc-file-systems.

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+
+The data itself is subdivided in blocks of size `<blocksize>` and is stored by Object Storage Servers (OSS) in one or more Object Storage Targets (OST).
+The blocksize and number of OSTs to use is defined by the striping settings. A new file or directory ihnerits them from its parent directory. The `lfs getstripe <path>` command can be used to get information on the actual stripe settings. For directories and empty files `lfs setstripe --stripe-count <count> --stripe-size <size> <directory/file>` can be used to set the layout. The simplest way to have the correct layout is to copy to a directory with the correct layout
```
msimberg, line 9:
```suggestion
The blocksize and number of OSTs to use is defined by the striping settings.
A new file or directory ihnerits them from its parent directory. The `lfs getstripe <path>` command can be used to get information on the actual stripe settings.
For directories and empty files `lfs setstripe --stripe-count <count> --stripe-size <size> <directory/file>` can be used to set the layout.
The simplest way to have the correct layout is to copy to a directory with the correct layout.
```

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+
+A blocksize of 4MB gives good throughput, without being overly big, so it is a good choice when reading a file sequentially or in large chuncks, but if one reads shorter chuncks in random order it might be better to reduce the size, the performance will be smaller, but the performance of your application might actually increase.
```
msimberg, line 11:
```suggestion
A blocksize of 4MB gives good throughput, without being overly big, so it is a good choice when reading a file sequentially or in large chunks, but if one reads shorter chunks in random order it might be better to reduce the size, the performance will be smaller, but the performance of your application might actually increase.
```

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+    lfs setstripe --stripe-count -1 --stripe-size 4M <big_files_dir>`
+    ```
+
+Lustre also supports composite layouts, switching from one layout to another at a given size `--component-end` (`-E`).
+With it it is possible to create a Progressive file layout switching `--stripe-count` (`-c`), `--stripe-size` (`-S`), so that fewer locks are required for smaller files, but load is distributed for larger files.
+
+!!! example "Good default settings"
+    ```console
+    lfs setstripe -E 4M -c 1 -E 64M -c 4 -E -1 -c -1 -S 4M <base_dir>
+    ```
```
msimberg, line 25: General question: do we know the defaults on capstor/ioppstor? It might be useful for users to know when they should actually care about changing the settings. Or are these "Good default settings" actually our defaults?

docs/guides/lustre-tuning.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Iopsstor vs Capstor
+
+`iopsstor` uses SSD as OST, thus random access is quick, and the performance of the single OST is high. `capstor` on another hand uses harddisks, it has a larger capacity, and  it also have many more OSS, thus the total bandwidth is larger.
```
msimberg, line 29:
```suggestion
Iopsstor uses SSD as OST, thus random access is quick, and the performance of the single OST is high. Capstor on another hand uses harddisks, it has a larger capacity, and  it also have many more OSS, thus the total bandwidth is larger.
```

Try to capitalize Capstor and Iopsstor for consistency with other pages. Unless you're talking about the actual paths that are mounted, I would avoid using `/capstor`.

msimberg (changes requested):
Thanks @fawzi, this is very nice to have.

It might be a good idea to at least cross-link (i.e. possibly both directions) with https://docs.cscs.ch/guides/storage/, https://docs.cscs.ch/storage/filesystems/, and https://docs.cscs.ch/alps/storage/ for better discoverability? The platform pages (e.g. https://docs.cscs.ch/platforms/hpcp/#file-systems-and-storage) also have some storage related content.

The info that you're adding is unique and new, but we might have to globally optimize a bit how we group all storage-related info.

docs/guides/lustre-tuning.md
fawzi: Changed references, merged with storage guide, and added some cross references.

docs/guides/lustre-tuning.md
fawzi: The performance of the metadata part *is* roughly the same on iopsstor and capstor, it depends on the number of metadata servers and users adding load to it. For a while backup was putting extra load on capstor. Datawriting of many small files might indeed be slightly better on iopsstor because SSD are faster

docs/guides/lustre-tuning.md
fawzi: The definition means first 4M use just one OST (less locks, less overhead), until 64M 4 OST, then all OST and 4MB blocksize, I tested it, but not heavily. I discussed already with @mpasserini , but he is not keen on changing the things especially as capstor might be replaced for MLP soon. I am willing to push users that are willing to set things, and if we then have more test coverage really change the default

bcumming (approved):
Thanks Fawzi.

These are a good starting point for us to develop more detailed docs over time.

I made a few tweaks and pushed them, to speed up the process.

msimberg (approved):

PR #178, opened 2025-07-02
Avoid port 6000 in pytorch

henrique:
optionally add warning somewhere?

### DIFF
--- a/docs/software/ml/pytorch.md
+++ b/docs/software/ml/pytorch.md
@@ -337,7 +337,7 @@ export OMP_NUM_THREADS=8 # (2)!
 # PyTorch environment variables #
 #################################
 export MASTER_ADDR=$(hostname) # (3)!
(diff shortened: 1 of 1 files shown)


docs/software/ml/pytorch.md
```diff
@@ -337,7 +337,7 @@ export OMP_NUM_THREADS=8 # (2)!
 # PyTorch environment variables #
 #################################
 export MASTER_ADDR=$(hostname) # (3)!
-export MASTER_PORT=6000
+export MASTER_PORT=29500
```
bcumming, line 340: How about adding a comment to this line about why?

henrique: +1 @bcumming, wording? something like:
Please avoid common ports like 6000, 60000 or <1024 which may be used by another service and could cause a silent tcp socket timeout.
?

PR #204, opened 2025-07-17
Add doc for scorep

jgphpc:


docs/software/devtools/vihps.md
```diff
@@ -0,0 +1,138 @@
+[](){#ref-devtools-vihps}
+# VI-HPS tools
+
+The [VI-HPS](https://www.vi-hps.org/tools) Institute (Virtual Institute for High Productivity Supercomputing) provides tools that can assist developers of simulation codes to address their needs in performance analysis.
+
+## [Score-P](https://www.vi-hps.org/projects/score-p/overview/overview.html)
+
+![logo_scorep](../../images/devtools/vihps/logo_scorep.png){ width="5%"}
```
bcumming, line 8: Is the logo necessary? It is easier to read (and search) if the name of the package is in plain text instead of an inline image.

Images also increase the size of the repository, and we suggest only including them where they significantly improve the docs.

docs/software/devtools/vihps.md
```diff
@@ -0,0 +1,138 @@
...
+## [Cube and Scalasca](http://www.vi-hps.org/tools/scalasca.html)
+
+![logo_scalasca](../../images/devtools/vihps/logo_scalasca.png){ width="5%"}
+ support the performance optimization of parallel programs with a collection of scalable trace-based tools for in-depth analyses of concurrent behavior. The analysis identifies potential performance bottlenecks - in particular those concerning communication and synchronization - and offers guidance in exploring their causes.
+
+## Quickstart guide
+
+The VI-HPS uenv is named `scorep` and it can be loaded into your environment as explained here and in the [uenv documentation][ref-uenv].
+
+??? example "Finding and pulling available `scorep` versions"
```
bcumming, line 31: How about expanding these boxes? All of the information in them is neccesary/good for readers to see without having to click

docs/software/devtools/vihps.md
```diff
@@ -0,0 +1,138 @@
...
+
+    ```bash
```
bcumming, line 33: Use `console` in cases like this, to render commands and output correctly.
https://docs.cscs.ch/contributing/#__tabbed_8_1

docs/software/devtools/index.md
```diff
@@ -28,6 +28,6 @@ This ensures that computational resources are utilized to their fullest potentia
 Learning to analyze the performance of an applications effectively is crucial to build a deeper understanding of how your code interacts with the underlying hardware.
 In this section we introduce the various performance analysis solutions available at CSCS.
 
-* [Linaro Forge MAP][ref-devtools-map]
 * [NVIDIA Nsight Developer Tools][ref-devtools-nsight]
-
+* [Linaro Forge MAP][ref-devtools-map]
+* [VI-HPS Tools][ref-devtools-vihps]
```
bcumming, line 33: Don't forget to add the vihps.md to the `mkdocs.yaml` file so that it appears in the main index.

bcumming (changes requested):
Thanks JG!

I have a few nit picks.

The main one is that the images of the product logos are not needed, and would increase the size of the repository by 230kb.

docs/software/devtools/vihps.md
jgphpc: we are talking about less than 250K here

bcumming: It all adds up over time.

But the main point is that using a picture where the name could be written in plain text makes the docs less readable and searchable.

docs/software/devtools/vihps.md
jgphpc: all the sciapps/*.md use bash

bcumming (approved):

PR #206, opened 2025-07-18
add a guide for creating custom uenv environments

bcumming:
In response to https://jira.cscs.ch/browse/SD-66316

https://docs.tds.cscs.ch/206/software/uenv/#custom-environments

### DIFF
--- a/docs/guides/terminal.md
+++ b/docs/guides/terminal.md
@@ -111,7 +111,7 @@ It is strongly recommended that customization in `~/.bashrc` should be kept to t
 
     If there is a critical error, like calling `exit`, the approach above won't work.
     In such cases, the only solution that doesn't require root permissions is to log in and hit `<ctrl-c>` during the log in.
-    With luck, this will cancel before the login process before `~/.bashrc` is executed, and you will be able to edit and fix `~/.bashrc`
+    With luck, this will cancel the login process before `~/.bashrc` is executed, and you will be able to edit and fix `~/.bashrc`.
     Note that you might have to try a few times to get the timing right.
 
     If this does not work, create a [service desk ticket][ref-get-in-touch] with the following message:


docs/guides/terminal.md
```diff
@@ -76,3 +76,49 @@ export PATH=$xdgbase/bin:$PATH
...
+
+    The first thing to try is to execute a command that will back up `~/.bashrc`, and remove `~/.bashrc`:
+    ```bash
+    ssh eiger.cscs.ch 'bash --norc --noprofile -c "mv ~/.bashrc ~/.bashrc.back"'
+    ```
+    If this works, you can then log in normally, and edit the backup and copy it back to `~/.bashrc`.
+
+    If there is a critical error, like calling `exit`, the approach above won't work.
+    In such cases, the only solution that doesn't require root permissions is to log in and hit `<ctrl-c>` during the log in.
+    With luck, this will cancel before the login process before `~/.bashrc` is executed, and you will be able to edit and fix `~/.bashrc`
```
RMeli, line 114:
```suggestion
    With luck, this will cancel the login process before `~/.bashrc` is executed, and you will be able to edit and fix `~/.bashrc`.
```

docs/software/uenv/index.md
```diff
@@ -601,3 +601,66 @@ uenv image find @'*'%gh200
 
 !!! note
     The wild card `*` used for "all systems" must always be escaped in single quotes: `@'*'`.
+
+[](){#ref-uenv-customenv}
+## Custom environments
+
+!!! warning "[Keep bashrc clean][ref-guides-terminal-bashrc]"
```
RMeli, line 608: Is this supposed to be a standalone warning (i.e. no indentation afterwards)?

RMeli (approved):

docs/software/uenv/index.md
bcumming: yes

PR #213, opened 2025-08-12
Update index.md

jpcoles-cscs:
Clarify that the value of SCRATCH discussed here is for the ML platform to avoid confusion if someone finds this through another link and doesn't notice it is only for the MLp.

### DIFF
--- a/docs/platforms/mlp/index.md
+++ b/docs/platforms/mlp/index.md
@@ -51,7 +51,7 @@ Scratch filesystems provide temporary storage for high-performance I/O for execu
 Use scratch to store datasets that will be accessed by jobs, and for job output.
 Scratch is per user - each user gets separate scratch path and quota.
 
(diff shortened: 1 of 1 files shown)


finkandreas (commented):
This is not really reflecting reality.

`bristen` is part of the ML platform but has `capstor` as SCRATCH defined. I think this applies purely to `clariden`, and not the ML platform as a whole (as of now)

jpcoles-cscs:
Hm, ok. Then this is even more confusing.

bcumming:
Note that this is in the MLP documentation, and it reflects reality on the platform.

The main docs for storage mention the difference between MLP and other platforms:

https://docs.cscs.ch/storage/filesystems/#scratch

I wonder if something changed on Bristen? Did SCRATCH point to iopstor in the past?

PR #216, opened 2025-08-12
Document a known GPU out-of-memory bug caused by file caches residing in GPU memory.

jpcoles-cscs:


docs/running/known-issues.md
```diff
@@ -0,0 +1,12 @@
+[](){#ref-known-issues}
+# Known Issues
+
+## Out of Memory on GH200 nodes
+There is a known issue with Nvidia GPU driver version R550.54.15, currently installed on all GH200 nodes, that can reduce the amount of available GPU memory.
```
bcumming, line 5: Is it possible to quote / demonstrate typical error messages that are created when the job fails with the error?
This is important so that this information can be found by search, and to help users (and CSCS staff in analysing a bug report).

bcumming (changes requested):

PR #209, opened 2025-07-22
MLP tutorials update

lukasgd:
This updates the existing MLP tutorials, to be merged with ML software in a next step. Also for the future, a Megatron-LM tutorial with recommended settings is left.

PR #231, opened 2025-08-19
Move MLP tutorials under software, add CE section to Pytorch including best practice for large-scale training

lukasgd:


docs/build-install/containers.md
```diff
@@ -43,11 +47,33 @@ podman build -t <image:tag> .
...
 ## Importing images in the Container Engine
 
 An image built using Podman can be easily imported as a squashfs archive in order to be used with our Container Engine solution.
 It is important to keep in mind that the import has to take place in the same job allocation where the image creation took place, otherwise the image is lost due to the temporary nature of `/dev/shm`.
 
+!!! warning "Preliminary configuration: Lustre settings for container images"
+    Since container images are large files and the filesystem is a shared resource, you need to configure the target directory according to [best practices for Lustre][ref-guides-storage-lustre] before importing the container image so it will be properly distributed across storage nodes.
+
+    ```bash
+    lfs setstripe -E 4M -c 1 -E 64M -c 4 -E -1 -c -1 -S 4M <path to image directory> # (1)!
```
henrique, line 71: I think it is not a good a idea to duplicate the command as you already linked it above. Also 64MB still seems a bit little for full striping, doesn't it?

docs/software/ml/pytorch.md
```diff
@@ -1,6 +1,207 @@
...
+3. (optionally) extending by a virtual environment
+4. submitting jobs with CE in SLURM 
+
+These steps are illustrated in the [machine learning platform tutorials][ref-software-ml-tutorials] and the instructions detailed in the [podman build guide][ref-build-containers].
+
+!!! info "Preliminary steps"
+    Before proceeding with the next steps, make sure you have storage for podman configured as in the [build guide][ref-build-containers-configure-podman] and make sure to apply [recommended Lustre settings][ref-guides-storage-lustre] to every directory (e.g. `$SCRATCH/ce-images`) dedicated to container images before importing them with enroot. This is necessary to guarantee good filesystem performance.
+
+    ```bash
+    lfs setstripe -E 4M -c 1 -E 64M -c 4 -E -1 -c -1 -S 4M $SCRATCH/ce-images # (1)!
```
henrique, line 24: same here:
I think it is not a good a idea to duplicate the command as you already linked it above. Also 64MB still seems a bit little for full striping, doesn't it?

docs/software/ml/tutorials/llm-nanotron-training.md
```diff
@@ -1,17 +1,20 @@
...
-It is recommended to follow the previous two tutorials on [LLM Inference][ref-mlp-llm-inference-tutorial] and [LLM Fine-tuning][ref-mlp-llm-fine-tuning-tutorial] first, as this will build upon them.
+It is recommended to follow the previous two tutorials on [LLM Inference][software-ml-llm-inference-tutorial] and [LLM Fine-tuning][software-ml-llm-fine-tuning-tutorial] first, as this will build upon them.
 
 ### Set up Podman
 
-If not already done as part of the [LLM Inference tutorial][ref-mlp-llm-inference-tutorial], edit your podman configuration in `$HOME/.config/containers/storage.conf` as follows:
+If not already done as part of the [LLM Inference tutorial][software-ml-llm-inference-tutorial], edit your podman configuration in `$HOME/.config/containers/storage.conf` as follows:
 
 ```toml title="$HOME/.config/containers/storage.conf"
 [storage]
```
boeschf, line 20: this path is different if you use XDG variables

bcumming (changes requested):
The software section is becoming quite large, with a lot of content.
We are trying to keep the material there quite generic.
The original location, in the guides section, was more appropriate.

docs/software/ml/pytorch.md
```diff
@@ -1,6 +1,207 @@
...
+TORCH_NCCL_ASYNC_ERROR_HANDLING = "1" # (7)!
+MPICH_GPU_SUPPORT_ENABLED = "0" # (8)!
+```
+
+1. It is important to use curly braces for environment variables used in the EDF
+2. The path `/users` is not mounted as a whole since it often contains user-specific initialization scripts for the host environment and many frameworks leave temporary data behind that can lead to non-trivial runtime errors when swapping container images. Thus, it is recommended to selectively mount specific subfolders under `${HOME}` if needed.
+3. You can use `${PWD}` as an alternative to use the path submitted from when the container is started
+4. This enables NCCL installed in the container to make effective use of the Slingshot interconnect on Alps by interfacing with the [AWS OFI NCCL plugin][ref-ce-aws-ofi-hook] with libfabric. While not strictly needed for single node workloads, it is good practice to keep it always on.
+5. This makes NCCL output debug info during initialization, which can be useful to spot communication-related issues in a distributed scenario (see later tutorials). Subsystems with debug log can be configured with `NCCL_DEBUG_SUBSYS`.
+6. Disable CUDA JIT cache
```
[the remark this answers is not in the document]
lukasgd: Ok, including your explanation from the uenv section in the next commit.

docs/software/ml/pytorch.md
```diff
@@ -374,16 +580,15 @@ srun bash -c "
...
-7. Avoid writing JITed binaries to the (distributed) file system, which could lead to performance issues.
-8. These variables should always be set for correctness and optimal performance when using NCCL, see [the detailed explanation][ref-communication-nccl].
-9. `RANK` and `LOCAL_RANK` are set per-process by the Slurm job launcher.
-10. Activate the virtual environment created on top of the uenv (if any).
+3. Enable more graceful exception handling, see [PyTorch documentation](https://pytorch.org/docs/stable/torch_nccl_environment_variables.html)
+4. Set the Triton home to a local path (e.g. `/dev/shm`) to avoid writing to the (distributed) file system.
+   This is important for performance, as writing to the Lustre file system can be slow due to the amount of small files and potentially many processes accessing it. Avoid this setting with the container engine as it may lead to errors related to mount settings of `/dev/shm` (use a filesystem path inside the container instead).
```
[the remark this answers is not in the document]
lukasgd: You mean `export TRITON_HOME=/dev/shm/`? That is exactly discouraged as mentioned in this paragraph.

> Avoid this setting with the container engine as it may lead to errors related to mount settings of `/dev/shm` (use a filesystem path inside the container instead).

It's also mentioned in the CE section that mounting of directories under `$HOME` should be done selectively. This is a learning from some non-trivial errors in large-scale runs.

docs/software/ml/tutorials/llm-nanotron-training.md
lukasgd: Will be included in next commit.

lukasgd:
The last commit should address your review comments @boeschf. @henrique, I think we need to work out a good Lustre default (in a ticket) and until then I would leave the slightly redundant but consistent recommendation in place (easy to search and replace). 

@bcumming, I'm interested in concrete feedback to improve the ML docs - please share if you have any ideas. For the merging of the MLP tutorials with the software section - this was previously announced in https://github.com/eth-cscs/cscs-docs/pull/209#issue-3253536463 and coordinated with @msimberg.

PyTorch is the main workhorse for most of Clariden's users, and as such deserves detailed docs I believe, which includes these tutorials. It can't all fit on a single page (like for some of the other scientific packages that also contain specific examples btw) as it would be overwhelming.

Currently, we don't have usage tracking for the docs, but given the support requests/Slack channel messages, it is my impression that many users are not sufficiently aware of these tutorials. A large factor contributing to this is their current placement on the website. We would like the users, though, to go through these materials as a first step and this is most naturally done when they are found right next to the framework documentation, which has the more advanced instructions.

I hope that's understandable, even though, not quite in line with your suggestion? As mentioned above, if you have concrete suggestions on how to improve the ML docs, that would be very welcome, please share them with us.

bcumming:
Discoverability of material like tutorials is something that we want to keep improving. However, putting more material into the Software section is not the solution, because the software section is already becoming too complicated, making discoverability harder.

The MPL tutorial is really good content that we don't have much of: a detailed step by step guide that shows how to build a workflow starting from scratch. Along the way it touches many topics, including containers, storage, different clusters and Slurm. As such, it makes sense to have it in a category that is separate from these topics and can link to them.

The guides section was a decent compromise, but might not be ideal 

- create a tutorials section (similar to guides)
- move the content into the MLP under a tutorials section

To improve discoverability, we need users to be directed towards the docs. Maybe we could provide a link in message of the day on Clariden/Bristen, that points users towards a getting started page that directly links to the tutorial?

docs/software/ml/index.md
```diff
@@ -2,22 +2,33 @@
 # Machine learning applications and frameworks
 
 CSCS supports a wide range of machine learning (ML) applications and frameworks on its systems.
-Most ML workloads are containerized to ensure portability, reproducibility, and ease of use across environments.
+Most ML workloads are containerized to ensure portability, reproducibility, and ease of use across machines.
```
teojgo, line 5:
```suggestion
Most ML workloads are containerized to ensure portability, reproducibility, and ease of use across systems.
```

docs/software/ml/pytorch.md
```diff
@@ -1,15 +1,250 @@
...
+
+1. Allocate an interactive session on a compute node
+2. Create a virtual environment on top of the existing Python installation in the container (only necessary the first time)
+3. Activate the newly created virtual environment (always necessary when running a Slurm job)
+4. Install additional packages (only run this from a single process to avoid race conditions)
+
+The changes made to the virtual environment will outlive the container as they are persisted on the distributed filesystem.
+
+!!! note
+    Keep in mind that
```
teojgo, line 129:
```suggestion
    Keep in mind that:
```

docs/software/ml/tutorials/index.md
```diff
@@ -0,0 +1,13 @@
+[](){#ref-software-ml-tutorials}
+# Machine Learning Platform Tutorials
+
+The LLM tutorials gradually introduce key concepts of the Machine Learning Platform in a series of hands-on examples. A particular focus is on the [Container Engine][ref-container-engine] for managing the runtime environment.
+
+In a [first tutorial][software-ml-llm-inference-tutorial], you will learn how to run inference with a LLM on a single node using a container from the NVIDIA GPU Cloud (NGC). Concepts such as container environment description, layering a thin virtual environment on top of the container image, and job launching and monitoring will be introduced.
```
teojgo, line 6:
```suggestion
In the [first tutorial][software-ml-llm-inference-tutorial], you will learn how to run inference with a LLM on a single node using a container from the NVIDIA GPU Cloud (NGC). Concepts such as container environment description, layering a thin virtual environment on top of the container image, and job launching/monitoring will be introduced.
```

jpcoles-cscs:
Here's my view on this:
The content is really good and we should get this out quickly. We can rework it once people start using it.
I would keep the tutorials where they were in Guides but rename the section to "Machine Learning" rather than "MLP Tutorials". There needs to be clear separation between reference docs and guide/tutorials. Otherwise it will become too cluttered. Furthermore it buries the guides too deep, whereas they should be quickly locatable. If you want, you could have a link from the PyTorch docs to the guides.

teojgo (changes requested):

docs/build-install/containers.md
henrique: could you remove the repetition for now please? so we only have 1 place to change after, and we can add it back once that is done...

bcumming: I agree with Henrique.
I have updated the docs to:
```
To ensure good performance for jobs on multiple nodes, take the time to configure the target directory using `lfs setstripe` according to [best practices for Lustre][ref-guides-storage-lustre] before importing the container image, or using `lfs migrate` to fix files that are already imported.
```
This makes the commands explicit, but let's us provide guidance on specific flags in one location.

docs/platforms/mlp/index.md
```diff
@@ -91,4 +91,4 @@ Project is per project - each project gets a project folder with project-specifi
 
 ## Guides and tutorials
 
-Tutorials for fine-tuning and running inference of LLMs as well as training an LLM with Nanotron can be found in the [MLP Tutorials][ref-guides-mlp-tutorials] page.
+Tutorials on how to set up and configure a machine learning environment in order to run LLM workloads such as inference, fine-tuning and multi-node training can be found in the [tutorials section][ref-software-ml-tutorials].
```
bcumming, line 94: I move this to the top of the platform page.

bcumming (approved):

PR #233, opened 2025-08-21
Add tip on how to check that the CXI provider is working correctly

RMeli:


docs/software/container-engine/resource-hook.md
```diff
@@ -169,6 +169,32 @@ The hook is activated by setting the `com.hooks.cxi.enabled` annotation, which
     4194304             23925.61
     ```
 
+!!! tip "How to check if the CXI provider is available"
```
bcumming, line 172:
```suggestion
!!! tip "How to check the CXI provider inside a container"
```

docs/software/container-engine/resource-hook.md
```diff
@@ -169,6 +169,32 @@ The hook is activated by setting the `com.hooks.cxi.enabled` annotation, which
...
+
+    You can check if the CXI provider is available using `fi_info`
+    (within the container):
```
bcumming, line 175: The `within the container` is the key context here (I think?)

Maybe start the section with something like "If you are uncertain about whether the CXI provider was configured correctly inside your running container..."

bcumming (changes requested):

PR #238, opened 2025-08-22
New pull request with whitelist words and switching to American English

cerlane:


docs/software/gssr/quickstart.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Example
+
+If you are submitting a batch job and the command you are executing is:
+
+    srun python test.py
+
+The corresponding srun command should be modified as follows.:
+
+    srun gssr profile -wrap="python abc.py"
```
finkandreas, line 36: above it it `test.py`, here it is `abc.py`. Should be consistent.

cerlane: Thanks Andreas, good point about `test.py` and `abc.py`. Updated. 

First of all, I will tell the user that people don't do whitespaces in file names on Linux. Jokes aside, I think the usual Linux applies. It should become `--wrap="python with\ whitespace.py"`. That is my assumption. Not tested though.

mkdocs.yml
```diff
@@ -82,6 +82,10 @@ nav:
       - 'Building uenv': software/uenv/build.md
       - 'Deploying uenv': software/uenv/deploy.md
       - 'Release notes': software/uenv/release-notes.md
+    - 'gssr':
```
bcumming, line 85: The "debugging and performance" section might be a good home for this?
* it adds some more context for readers looking for tools
* it avoids too many top level categories

docs/software/gssr/containers.md
```diff
@@ -0,0 +1,120 @@
+[](){#ref-gssr-containers}
+# gssr - Containers Guide
+
+CSCS highly recommends that all users leverage on container solutions on our Alps platforms so as to flexibly configure any required user environments of their choice within the containers. Users thus have maximum flexibility as they are not tied to any specific operating systems and/or software stacks.
```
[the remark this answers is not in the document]
cerlane: Out of curiosity, do we even have such a guidance page? Fair statement about uenv and container though. I am mainly focus with ML users and they were told to do containers so I forget about uenv. I can remove that.

mkdocs.yml
cerlane: Yes, that might be a good place too. The original home suggestion came after a discussion with Nina. Simply because the category we belong to is profiling, which can be under performance but not completely. I can accept being there. I will move. Thanks.

.github/actions/spelling/allow.txt
```diff
@@ -241,6 +246,7 @@ kubeconfig
 ceph
 rwx
 rwo
+sqsh
```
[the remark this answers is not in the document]
cerlane: let me see how i can improve this. It is a fair statement

PR #240, opened 2025-08-22
remove gordon bell guide now it is over

bcumming:


RMeli (commented):
Should the "Recommendations on run configuration" section stick around somewhere in the documentation as general recommendations for large scale runs?

msimberg (approved):
Thanks for rebasing this @bcumming. I'll look into integrating some of the tips and tricks into other sections.

PR #241, opened 2025-08-25
Fix/feedbacks

cerlane:


docs/software/devtools/gssr/containers.md
```diff
@@ -0,0 +1,120 @@
+[](){#ref-gssr-containers}
+# gssr - Containers Guide
+
+The following guide will explain how to install and use `gssr` within a container.
+
+Most CSCS users leverage on the base containers with pre-installed CUDA from Nvidia. As such, in the following documentation, we will use a PyTorch base container as an example.
```
bcumming, line 6:
```suggestion
Most CSCS users leverage on the base containers with pre-installed CUDA from Nvidia.
As such, in the following documentation, we will use a PyTorch base container as an example. 
```
Sorry if it wasn't clear in the previous review, but the "one sentence per line" rule looks like this suggested change.
You don't have to change the content at all, instead put each sentence in a paragraph on its own line: the generated docs will join them together into a paragraph (you need a blank line to start a new paragraph)

bcumming (changes requested):
Thanks for the contribution!

I have some suggested changes, and I have tried to add some extra information that might have been missing in earlier reviews.

PR #249, opened 2025-08-29
Update policies introducing maintenance policies

lucamar:
I provide an update of the policies as described in [SRM-274](https://jira.cscs.ch/browse/SRM-274) in view of the presentation "General Information for Users on Alps" at the UserLab Day 2025.


twrobinson (approved):
Looks good. The left had menu needs to be updated I think

PR #253, opened 2025-09-03
Add `NCCL_PROTO=^LL128` to recommended NCCL environment variables

msimberg:


RMeli (commented):
Can't comment on the performance, but the changes LGTM.

PR #265, opened 2025-09-15
CE: Add known issue about buffer overflow errors with long command trings

Madeeks:


docs/software/container-engine/known-issue.md
```diff
@@ -1,3 +1,9 @@
+## Buffer overflow errors with long command strings
+
+We are aware of an issue which is causing a buffer overflow error and abrupt termination of jobs using the CE when entering very long strings as the command to execute in the Slurm job step.
```
msimberg, line 3: How about adding an approximate date for when this started happening? E.g. "We are aware of an issue, as of the system update on 10th of September 2025, which is causing..." or something like that? Just to put some context to "we'll provide an update shortly".

Additionally, could you add an example of the error message (for searchability). If I understand correctly, it's
```
0: *** buffer overflow detected ***: terminated
```
?

Madeeks: Updated message with suggestions and progress made in the meantime

PR #268, opened 2025-09-25
📚 Build-uenv: consistently use `<build-path>`

mbercx:
The documentation section was using both `<build-path>` and `<build-dir>`. This makes it unclear that they are both referring to the same string.

Here we choose to consistently use `<build-path>`, since it is a path to a directory.

### DIFF
--- a/docs/build-install/uenv.md
+++ b/docs/build-install/uenv.md
@@ -198,7 +198,7 @@ you can provide it as an argument to `uenv-spack`:
 
 ### Build the software
 
-Once specs have been added to `spack.yaml`, you can build the image using the `build` script that was generated in `<build-dir>`:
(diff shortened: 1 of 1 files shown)


mbercx:
Not an expert on `uenv` (hence I'm reading the docs ^^), so feel free to close in case my assumption that `build-path` and `build-dir` are the same is incorrect.

bcumming (approved):
Thanks @mbercx for improving the docs!

 Technically, build-dir might be a little more appropriate, because in POSIX, path can be a directory, file, symlink etc.
 But, it doesn't matter whether we use build-dir or build-path, so long as it is consistent.

mbercx:
> Thanks @mbercx for improving the docs!

Very happy to contribute! Also: I'm back to using the CSCS facilities after a while, and I must say these new docs look spiffy. 👌

PR #287, opened 2025-10-22
Fix serve helper not reloading

bcumming:
force the serve helper to automatically reload to work around issue with click 8.3.0

### DIFF
--- a/serve
+++ b/serve
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 # use uv to run mkdocs with mkdocs-material and its dependencies installed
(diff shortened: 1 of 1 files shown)


msimberg (approved):
Thanks @bcumming!

I think this is good.

Just a note to anyone else who's like me and binds the server to a different port (the default conflicts with another service I have running locally), you'll have to manually add the `--livereload` flag, i.e. `./serve --livereload --dev-addr 127.0.0.1:<port>`. This is an edge case and I don't think we need to cover that with the defaults.

PR #303, opened 2025-11-27
update uenv release notes; uenv->uenvs for more than one uenv

bcumming:
This is a quick update that:

- updates the release notes for uenv with notes about 9.0.1, 9.1.0, 9.1.1
- cleans up references to some bugs since uenv has been upgraded to 9.1.0 on the main systems
- start converting the docs to use uenvs as the plural of uenv (seems like I was the only person on earth who thought that uenv should be the plural of uenv).

### DIFF
--- a/docs/software/uenv/build.md
+++ b/docs/software/uenv/build.md
@@ -1,9 +1,9 @@
 [](){#ref-uenv-build}
-# Building uenv
+# Building uenvs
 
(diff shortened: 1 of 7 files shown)


msimberg (commented):
Some minor nits and typos, looks good otherwise.

msimberg (approved):

PR #343, opened 2026-02-19
Clarification for Linaro DDT setup

RMeli:
As far as I can tell, the `activation` script was removed in `25.1`.


jgphpc (approved):

bcumming (changes requested):
I have made some changes that remove all mention of the `activate` script - this was a workaround of old limitations of uenv.

I would suggest making the following changes to the uenv recipe for `linaro-forge`

* do not include compilers in the uenv view (set the following line to false)
    - https://github.com/eth-cscs/alps-uenv/blob/main/recipes/linaro-forge/25.1/environments.yaml#L10
* name the uenv `linaro-forge` (the advice in the docs that you should run `uenv status` must have been giving error messages for months because of this).
    - https://github.com/eth-cscs/alps-uenv/blob/main/recipes/linaro-forge/25.1/config.yaml#L1
* name the main view `forge`, not `default` (to disambiguate)
    - https://github.com/eth-cscs/alps-uenv/blob/main/recipes/linaro-forge/25.1/environments.yaml#L7
* finally you can set the "default" view in `config.yaml`, which should remove the need to use views at all when the new version 10 of uenv is deployed:
    - https://eth-cscs.github.io/stackinator/recipes/#setting-default-views

Then removing old copies from the system, to make this documentation valid.

jgphpc:
we also need reframe.yaml to point to the new view

bcumming (approved):

PR #378, opened 2026-04-20
Scaffold of inference related documentation

schuups:
I am creating this PR primarily to get a rendered version for discussion with Pablo and team. 

Feel free to adapt to fit the overall guidelines of our docs.

https://cscs-docs-preview.svc.cscs.ch/378/services/inference


naevtamarkus:
Adding description of inference services, as well as having both Stefano and Pablo as entry points for service design.

bcumming:
I made a few tweaks to file names and link names.
The pattern for links is
`ref-inference` for the top level index
`ref-inference-api` for the L1 API service. All sub-links on that page should start with `ref-inference-api-`
Each other page in inference will have a top level link `ref-inference-X` where `X` is a short description of the service.

docs/services/inference/api.md
```diff
@@ -66,7 +66,7 @@ Approved projects are given an authentication token, which can be retrieved and
     Project key management is evolving. Future improvements include multiple keys per project, self-service rotation, revocation, scoped access, auditability, and quota or budget controls.
 
 !!! todo
-    Instead of adding a screenshot, can't we 
+    Add a screenshot to see how to obtain the API key
```
bcumming: does this need a screenshot?

And, is the method for obtaining API keys documented elsewhere in our docs? If it is, link to that (and improve those docs if needed). That way we document everything once.

schuups: @naevtamarkus I am aware of the process but I don't have access to it to take printscreens. Can you do this?

schuups: We should say that this is temporary implementation (e.g., the API key is managed in a strange way vs. standards... i.e. the key is visible (not hidden), and not revokable, etc...)

naevtamarkus: This is already explained in the limitations... I would avoid having "construction site" warnings all over the doc, but rather explain how it works "today" and change the doc when we change things... would you agree?

docs/services/inference/api.md
```diff
@@ -0,0 +1,131 @@
...
+## Access
+
+### Request access
+
+Early usage of this service requires an invitation.
+If you would like to participate, please contact Pablo Fernandez ([`pablo.fernandez@cscs.ch`](mailto:pablo.fernandez@cscs.ch)) describing your use case, relevant project or organizational context, and an estimate of your expected requirements including load, preferred models, and availability expectations.
+
+### Obtain your authentication token
+
+Approved projects are given an authentication token, which can be retrieved and managed through [project management portal][ref-account-waldur].
```
msimberg: The portal page currently has no info on how to retrieve and manage the token. Can that be added (here or on the portal page)?

docs/services/inference/api.md
```diff
@@ -0,0 +1,131 @@
...
+!!! example "List available models"
+    ```console
+    $ curl -X GET "https://ai-gateway.svc.cscs.ch/v1/models" \
+      -H "Authorization: Bearer <AUTHENTICATION_TOKEN>" \
+      -H "Content-Type: application/json"
+    ```
+
+!!! example "Chat completion request"
+    ```console
+    $ curl -X POST "https://ai-gateway.svc.cscs.ch/v1/chat/completions" \
```
msimberg: I don't know if this is temporary, but currently the url is:
```suggestion
    $ curl -X POST "https://llm-proxy.svc.cscs.ch/chat/completions" \
```
. At least the ai-gateway url doesn't work for me.

docs/services/inference/api.md
```diff
@@ -0,0 +1,131 @@
...
+      -H "Authorization: Bearer <AUTHENTICATION_TOKEN>" \
+      -H "Content-Type: application/json" \
+      -d '{
+        "model": "apertus-70b-instruct",
```
msimberg:
```suggestion
        "model": "Apertus-70B-Instruct-2509",
```

docs/services/inference/api.md
```diff
@@ -0,0 +1,131 @@
...
+      }'
+    ```
+
+!!! example "Claude Code CLI"
+    Example environment configuration to be set before starting a `claude` session.
+    ```console
+    $ export ANTHROPIC_API_KEY=<AUTHENTICATION_TOKEN>
+    $ export ANTHROPIC_BASE_URL=https://ai-gateway.svc.cscs.ch/v1
+    $ export ANTHROPIC_MODEL=apertus-70b-instruct
+    ```
```
msimberg: This seems to not work currently work with the llm-proxy url, which is openai compatible. One option seems to be to use a litellm proxy locally. This may need documenting or better: the anthropic url, if it exists, should be documented.

docs/services/inference/api.md
schuups: The team confirmed yesterday that, for now, `llm-proxy.svc.cscs.ch` is the name to use. Could change again later, but for now ok to put in the docs.

docs/services/inference/api.md
schuups: There is a pending request from users, to name models after the exact same Hugging Face IDs. In this case, it would be `swiss-ai/Apertus-70B-Instruct-2509`. Until LiteLLM is in place, the names will be like per @msimberg suggestion.

docs/services/inference/api.md
schuups: Added illustrative image for this purpose

docs/services/inference/api.md
schuups:
```
!!! todo
    Describe API support.
    If we provide both OpenAI and Anthropic APIs, is it sufficient to provide links to external documentation for these APIs, with notes about any differences?
```


Once we fix the endpoint to support both, getting started examples in our doc should be succinct and we can include links to external docs for common coding agents.

bcumming (approved):

PR #383, opened 2026-04-22
Update slurm.md to add nodes=1 to guarantee a specific node

twrobinson:
nodelist option by itself is only a request not a guarantee. To guarantee the specific nodes you need to add the nodes=1 flag in addition.


gsavva:
thanks @twrobinson for this PR!
I confirm that a command like
```
srun --jobid=<JOBID> --nodes=1 --nodelist=nidXXXXXX --overlap --pty bash
```
allows me to connect to a specific compute node.

bcumming (approved):

PR #390, opened 2026-05-11
Documentation for CP2K node sharing

abussy:
Updated the CP2K documentation with an example of node sharing. 4 CP2K calculations are launched simultaneously on the same node, each using a single GPU and the 72 adjacent CPU cores. I tested it thoroughly with `jobreport` and the reported `CPU_MASK` to verify that it does what it claims.

A couple of open questions:
1) Here I add a `select_gpu.sh` script, which starts the MPS daemon on a given GPU. It is very close to `mps-wrapper.sh`, so I wonder whether we should modify the latter to accept a GPU id as argument, rather than increasing the number of bash scripts scattered around?
2) I put the example in the CP2K docs, because it is rather specific. However, it could easily be generalized and put in the Slurm docs instead. I like the current solution, personally.

### DIFF
--- a/docs/software/sciapps/cp2k.md
+++ b/docs/software/sciapps/cp2k.md
@@ -166,6 +166,169 @@ sbatch run_cp2k.sh
          per node. Experiments have shown that CP2K performs and scales better when the number of MPI ranks is a power
          of 2, even if some cores are left idling. 
 
+[](){#ref-4-cp2k-calculation-1-node}
+#### Running multiple CP2K calculations simultaneously on a single Daint node
+GH200 nodes offer a very large amount of computing resources, and some calculations are simply too small to make use of these resources efficiently. A symptom of that is when running on 2 nodes is barely faster, or even slower, than a single node. If the scientific investigation permits, up to 4 independent CP2K calculations can run in parallel on a single node, each using 72 CPU cores and a GPU. This could typically be the case when running various MD trajectories at different temperatures, or with different starting configurations. For optimal resource utilization, the 4 calculations should have similar wall times. The `H2O-128.inp` benchmark from the CP2K [git repository](https://github.com/cp2k/cp2k/blob/support/v2026.1/benchmarks/QS/H2O-128.inp) is one such an example (the wall time on 1 and 2 nodes is roughly equivalent). Below is a Slurm submission script launching 4 `H2O-128.inp` benchmark calculations simultaneously:
+
+??? example "Running 4 independent CP2K calculations on a single node"
+    ```bash title="run_cp2k_on_split_node.sh"
+    #!/bin/bash -l
+
+    #SBATCH --job-name=cp2k-job
+    #SBATCH --time=00:30:00
+    #SBATCH --nodes=1
+    #SBATCH --account=<ACCOUNT>
+    #SBATCH --hint=nomultithread
+    #SBATCH --no-requeue
+    #SBATCH --uenv=<CP2K_UENV>
+    #SBATCH --view=cp2k
+
+    # Set the number of MPI ranks per calculation (one calculation per GPU, 4 in total).
+    # The number of OMP threads is automatically inferred.
+    nranks_per_calc=8
+    if !((nranks_per_calc > 0 && 72 % nranks_per_calc == 0)); then
+        echo "Error: the number of ranks per calculation must be an integer divisor of 72 (# CPU cores per GPU)" >&2
+        exit 1
+    fi
+    ncpus=$((288/$nranks_per_calc))
+
+    # Usual CP2K environment variables
+    export CUDA_CACHE_PATH="/dev/shm/$RANDOM"
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export MPICH_MALLOC_FALLBACK=1
+    export OMP_NUM_THREADS=$((72/$nranks_per_calc - 1))
+    ulimit -s unlimited
+
+    # Kill running mps daemon, if any. Usually done in mps-wrapper.sh
+    pkill --uid $(id -un) '^nvidia-cuda-mps-' || true
+
+    # Important note: with the --overlap option, each srun job step will share resources. Therefore, we allocate all resources
+    # for each job, i.e. nranks_per_calc*ncpus==288, and distribute the work with different CPU and GPU masks.
+    if !((nranks_per_calc * ncpus == 288)); then
+        echo "Error: all resources MUST be allocated for each calculation" >&2
+        exit 1
+    fi
+
+    # CPU masks to ensure each calculation runs on a different set of CPU cores, close to a given GPU.
+    mask0="mask_cpu:0xffffffffffffffffff"                                                        # cores 0-71
+    mask1="mask_cpu:0xffffffffffffffffff000000000000000000"                                      # cores 72-143
+    mask2="mask_cpu:0xffffffffffffffffff000000000000000000000000000000000000"                    # cores 144-215
+    mask3="mask_cpu:0xffffffffffffffffff000000000000000000000000000000000000000000000000000000"  # cores 216-287
+
+    # First calculation on GPU 0 and CPU cores 0-71
+    srun -u --overlap \
+         --cpu-bind=$mask0,verbose \
+         --ntasks=$nranks_per_calc \
+         --cpus-per-task=$ncpus \
+         --output="output0.out" \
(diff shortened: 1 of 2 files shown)


docs/software/sciapps/cp2k.md
```diff
@@ -162,6 +162,169 @@ sbatch run_cp2k.sh
...
+    # The number of OMP threads is automatically inferred.
+    nranks_per_calc=8
+    if !((nranks_per_calc > 0 && 72 % nranks_per_calc == 0)); then
+        echo "Error: the number of ranks per calculation must be an integer divisor of 72 (# CPU cores per GPU)" >&2
+        exit 1
+    fi
+    ncpus=$((288/$nranks_per_calc))
+
+    # Usual CP2K environment variables
+    export CUDA_CACHE_PATH="/dev/shm/$RANDOM"
```
msimberg:
```suggestion
    export CUDA_CACHE_PATH="/dev/shm/$USER/cuda_cache"
```
to align with what we already set on daint and otherwise in the CP2K docs.

docs/software/sciapps/cp2k.md
```diff
@@ -162,6 +162,169 @@ sbatch run_cp2k.sh
...
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export MPICH_MALLOC_FALLBACK=1
+    export OMP_NUM_THREADS=$((72/$nranks_per_calc - 1))
+    ulimit -s unlimited
+
+    # Kill running mps daemon, if any. Usually done in mps-wrapper.sh
+    pkill --uid $(id -un) '^nvidia-cuda-mps-' || true
```
msimberg: This assumes the script is only ever run on at most one node. Maybe this can error out if run with more than one node?

docs/software/sciapps/cp2k.md
```diff
@@ -162,6 +162,169 @@ sbatch run_cp2k.sh
...
+    ncpus=$((288/$nranks_per_calc))
+
+    # Usual CP2K environment variables
+    export CUDA_CACHE_PATH="/dev/shm/$RANDOM"
+    export MPICH_GPU_SUPPORT_ENABLED=1
+    export MPICH_MALLOC_FALLBACK=1
+    export OMP_NUM_THREADS=$((72/$nranks_per_calc - 1))
+    ulimit -s unlimited
+
+    # Kill running mps daemon, if any. Usually done in mps-wrapper.sh
```
msimberg:
```suggestion
    # Kill running MPS daemon, if any. Usually done in mps-wrapper.sh
```

docs/software/sciapps/cp2k.md
```diff
@@ -162,6 +162,169 @@ sbatch run_cp2k.sh
...
+    }
+
+    # First argument of select_gpu.sh is the selected GPU id (integer from 0 to 4)
+    if [[ "$1" =~ ^[0-3]$ ]]; then
+        GPU_ID="$1"
+        shift
+    else
+        echo "Error: invalid GPU id (must be an integer in [0,3])" >&2
+        exit 1
+    fi
```
msimberg: Why not use `numa_node`?

msimberg (commented):
Sorry for the delay in giving feedback.

> A couple of open questions:
> - Here I add a select_gpu.sh script, which starts the MPS daemon on a given GPU. It is very close to mps-wrapper.sh, so I wonder whether we should modify the latter to accept a GPU id as argument, rather than increasing the number of bash scripts scattered around?

I agree this would be nice. If you have motivation to look a bit more into this I think it wouldn't require big changes to make that work. Currently the generic mps-wrapper.sh script assumes that all four GPUs are used by the job, but this doesn't work in the case you're introducing and it doesn't work e.g. on the shared partition on santis. I think the most generic would be if we can infer which numa nodes are used by the whole job and only start the MPS daemon for those numa nodes. This would cover both the current mps-wrapper.sh use case, the shared partition, and the use case on this PR. And it wouldn't require setting GPUs explicitly. However, if that becomes too complicated, maybe just allowing to override the GPU(s) explicitly is the most straightforward solution.

> - I put the example in the CP2K docs, because it is rather specific. However, it could easily be generalized and put in the Slurm docs instead. I like the current solution, personally.

I think we can probably stick to CP2K for now, and generalize later as/when we get more examples of this pattern?

docs/software/sciapps/cp2k.md
abussy: Single node is indeed the assumption, but ideally this would stay within `mps-wrapper.sh`. I'll see if I can generalize it.

abussy:
For later reference, here is my current best idea:

Run the `mps-wrapper.sh` first, starting a MPS daemon on each GPU, then then launch each job step with the proper `CUDA_VISIBLE_DEVICES` and MPS log paths. It would look something like that:


```
srun ./mps-wrappers.sh bash -c "echo 'Setting up MPS daemons...'"          # using a universal MPS script

# First calculation on GPU 0 and CPU cores 0-71
CUDA_VISIBLE_DEVICES=0 CUDA_MPS_PIPE_DIRECTOY="..." CUDA_MPS_LOG_DIRECTORY="..." \
     srun -u --overlap \
     --cpu-bind=$mask0,verbose \
     --ntasks=$nranks_per_calc \
     --cpus-per-task=$ncpus \
     --output="output0.out" \
     cp2k.psmp -i H2O-128.inp \
     & j0=$!

# Second calculation on GPU 1 and CPU cores 72-143
CUDA_VISIBLE_DEVICES=1 CUDA_MPS_PIPE_DIRECTOY="..." CUDA_MPS_LOG_DIRECTORY="..."  \
     srun -u --overlap \
     --cpu-bind=$mask1,verbose \
     --ntasks=$nranks_per_calc \
     --cpus-per-task=$ncpus \
     --output="output1.out" \
     cp2k.psmp -i H2O-128.inp \
     & j1=$!
     
...
```

This way, there is only a single version of `mps-wrapper.sh`. The oversubscription script becomes a bit more complicated, but I think it's the better option. I'll finalize that when I come back from holidays.

abussy:
I believe my last commit introduces a much better solution, where the standard `mps-wrapper.sh` script can be used out of the box.

docs/software/sciapps/cp2k.md
```diff
@@ -166,6 +166,109 @@ sbatch run_cp2k.sh
...
+    #SBATCH --job-name=cp2k-job
+    #SBATCH --time=00:30:00
+    #SBATCH --nodes=1
+    #SBATCH --account=<ACCOUNT>
+    #SBATCH --hint=nomultithread
+    #SBATCH --no-requeue
+    #SBATCH --uenv=<CP2K_UENV>
+    #SBATCH --view=cp2k
+
+    srun --ntasks=4 --cpu-bind=socket ./mps-wrapper.sh ./parallel_jobs.sh
```
msimberg, line 186: Just to check, because the behaviour is quite unintuitive, should/could this be:

```suggestion
    srun --ntasks=1 --cpu-bind=socket ./mps-wrapper.sh ./parallel_jobs.sh
```
?

The unintuitive part (to me at least) is that first four ranks are started with the first rank starting the MPS daemons. Then the first rank, and only the first rank, does a nested srun with multiple ranks per GPU. To me it's not immediately obvious that the nested sruns will run on the same allocation, but it may just be me. Do you think it'd be overkill to add a note that all sruns, including the nested ones, will still run on the same allocation?

msimberg (commented):
Just one clarifying question/comment. Very nice that you got this working with the original MPS wrapper!

docs/software/sciapps/cp2k.md
abussy: Concerning the first `srun` call, i.e.
```
srun --ntasks=4 --cpu-bind=socket ./mps-wrapper.sh ./parallel_jobs.sh
```
I decided to set 4 tasks with `--cpu-bind=socket` such that each rank is bound to a numa node. Within `mps-wrapper.sh`, there is some waiting time for a pid file to be created per numa node (meaning MPS daemon is ready). If run with a single rank, there is a risk that the subsequent calculations start before it is ready.

However, I now realize that I should start each calculation from a different rank, unless there is an implicit barrier (which I don't think there is). I'll change that.

> Do you think it'd be overkill to add a note that all sruns, including the nested ones, will still run on the same allocation?

Not overkill, I'll add it as well.

bcumming (approved):

msimberg (approved):

PR #392, opened 2026-05-18
Strip prompts when copying code snippets

rjanalik:
The copy button in top-right of the code snippet copies only the code without the command line prompt '$' and the command output.

Closes eth-cscs/cscs-key#6

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -462,7 +462,10 @@ Use [code blocks](https://squidfunk.github.io/mkdocs-material/reference/code-blo
 The documentation uses [pygments](https://pygments.org) for highlighting.
 See [list of available lexers](https://pygments.org/docs/lexers/#) for the languages that you can use for code blocks.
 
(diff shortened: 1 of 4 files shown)


bcumming (changes requested):
Can you please update the contributing guide (see where we document using `console`) to add this advice explicitly.

docs/contributing/index.md
```diff
@@ -462,7 +462,10 @@ Use [code blocks](https://squidfunk.github.io/mkdocs-material/reference/code-blo
 The documentation uses [pygments](https://pygments.org) for highlighting.
 See [list of available lexers](https://pygments.org/docs/lexers/#) for the languages that you can use for code blocks.
 
-Use [`console`](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer) for interactive sessions with prompt-output pairs:
+Use a session lexer for interactive sessions with prompt-output pairs:
```
bcumming: What does this mean? Can you replace this with plain English.
And explain _why_ we do this (so that the copy option works).
That way it is clear that this isn't a random rule.

bcumming: I see now that you mention this below - try to tie it all together.

bcumming (approved):

PR #403, opened 2026-06-03
Document CP2K slowdown

RMeli:
Suggest to switch off GPU-aware MPI if not needed.


msimberg (approved):
Looks good. Since @anfink promised to keep `/capstor/store/cscs/cscs/public/temp/cuptrgetattr_override.so` around for at least a while, we could also mention that workaround here (now or separate PR, I can also open one). It might also belong on the cray-mpich page with a link from here instead.

RMeli:
I think the `cray-mpich` page would be the best place, with a link from here. Is this something we now want to advertise more widely?

PR #406, opened 2026-06-09
Update inference docs with claude/opencode information and internal service

msimberg:
Moves the claude code section to a separate coding agent section. Adds instructions on setting up opencode.

Also adds a note on the internal inference service at the bottom of the inference page.

### DIFF
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -205,3 +205,10 @@ Once connected, you can choose models configured in the config.
 * Detailed self-service telemetry is limited today.
 * Documentation and model-specific configuration transparency are work in progress.
 * Load balancing and other QoS need to be understood.
+
+
+## CSCS-internal inference service
+
+For CSCS staff, an internal inference service is available on CSCS VPN through the https://ai-gateway.svc.cscs.ch URL.
+Usage is otherwise identical to the https://llm-proxy.svc.cscs.ch endpoint.
+For more information and keys, join the #cscs-inference-services Slack channel.
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -58,37 +58,60 @@ However, CSCS collects infrastructure metrics and telemetry, including prompt an
 [](){#ref-inference-api-quickstart}
 ## Quick Start
 
-Before using the API, obtain an authentication token by following the [access guide][ref-inference-api-access]. Include this token in every API request.
+Before using the API, obtain an authentication token by following the [access guide][ref-inference-api-access].
+Include this token in every API request.
 
-!!! example "Anthropic's `claude`"
-    Example environment configuration to be set before starting a `claude` session.
-    ```console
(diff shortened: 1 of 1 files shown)


docs/services/inference/api.md
```diff
@@ -122,6 +145,54 @@ The service is accessed through the gateway base URL `https://llm-proxy.svc.cscs
 [](){#ref-inference-api-guides}
 ## Guides
 
+### Setting up coding agents to use the inference service
```
bcumming, line 148: What about making this a fully fledged guide in the "guides" section of the docs:

https://docs.cscs.ch/guides/

The reasoning is that setting up a coding agent is not specific to the CSCS inference service, and will grow to cover other topics like sand boxing etc.

docs/services/inference/api.md
```diff
@@ -122,6 +145,54 @@ The service is accessed through the gateway base URL `https://llm-proxy.svc.cscs
...
+
+Below are instructions for setting up [Claude Code](https://claude.com/product/claude-code) and [OpenCode](https://opencode.ai).
+For more details and for other agents, see their respective documentation pages.
```
sekelle, line 151: Propose to add the following:

!!! info
Both agent frameworks can be launched inside a loaded UENV, which will allow the agents to build and run your project.

sekelle: From a login node, those agents will also have access to SLURM and are in principle able to submit jobs if properly instructed. This has to be handled very carefully; as a user you are responsible for the actions of the agents that you launch. Compute resources used by agent-launched SLURM jobs will be billed towards your project account.

sekelle: !!! todo
    Add some information about access restriction and containers.

sekelle (changes requested):

msimberg:
@sekelle and @bcumming I half-agree with both of you. I agree that there's a lot more that could fit in a generic "agents on alps" guide, and I can follow up with that. However, the changes in this PR specifically talk about configuring the inference service with claude code and opencode, which I think very much belongs with the inference service page.

I like all of your suggestions, and I think we may also need to include something about "user is still responsible for actions of an agent". I also don't really yet have best practices for constraining agents on alps, but this is something that we should provide guidance on. The inference page and agents page can also link to each other, but I think they are separate guides.

Would you be fine with me opening a follow up PR creating a separate guide for "agents on alps" with the points you've brought up?

sekelle:
We can drop the part about access restrictions and containers. Would you still object to keeping the first two additions? We can remove it once we have an "agents on alps" guide where we also talk about sandboxing etc.

msimberg:
> We can drop the part about access restrictions and containers. Would you still object to keeping the first two additions? We can remove it once we have an "agents on alps" guide where we also talk about sandboxing etc.

Not at all if it's just temporary.

sekelle (approved):

PR #413, opened 2026-06-12
CE: expand docs on netstack artifacts

Madeeks:
Netstack artifacts are becoming defaults on Clariden and Daint with the maintenance on June 17th.

Since they are no longer a fallback or optional feature, their docs coverage should be improved to help users understand this feature.

PR #414, opened 2026-06-12
CE: AWS OFI NCCL update

Madeeks:
Update content about AWS OFI NCCL plugin and containers, in particular:
- update plugin variants supported by the CE;
- Introduce and explain dynamically-linked plugins;
- Add known issue about dynamically-linked plugins with images missing generic `libcudart.so` links.

This PR requires #413 because it references a navigation label in the new Known Issue.

### DIFF
--- a/docs/software/communication/nccl.md
+++ b/docs/software/communication/nccl.md
@@ -56,18 +56,18 @@ To use NCCL in a container, enable the [AWS OFI hook][ref-ce-aws-ofi-hook] in th
 
 ```toml
 [env]
-PMIX_MCA_psec="native" # (1)!
+PMIX_MCA_psec="^munge" # (1)!
 
 [annotations]
 com.hooks.aws_ofi_nccl.enabled = "true"    # (2)!
-com.hooks.aws_ofi_nccl.variant = "cuda12"  # (3)!
+com.hooks.aws_ofi_nccl.variant = "cuda-dl"  # (3)!
 ```
 
-1. Ensures PMIx uses the same security domain as Slurm. Otherwise PMIx will print warnings at startup.
+1. Ensures PMIx does not use the Munge security domain, which is not exposed into containers. Otherwise PMIx will print warnings at startup.
 2. Enable the AWS OFI plugin.
-3. Take care to match the major CUDA version installed in the container.
+3. The dynamically-linked (`dl`) variant is generally recommended for portability across CUDA versions. Statically linked variants must match the major CUDA version installed in the container. More details [here][ref-ce-aws-ofi-hook].
 
-Because NCCL uses OpenMPI in the container to perform initial setup, which in turn uses [PMIx](https://pmix.org/) for wire-up, pass the `--mpi=pmix` option to `srun` when launching jobs.
+Because the NCCL Tests use OpenMPI in the container to perform initial setup, which in turn uses [PMIx](https://pmix.org/) for wire-up, pass the `--mpi=pmix` option to `srun` when launching jobs.
 
 ```console
 $ srun --mpi=pmix -n8 -N2 --environment=nccl-test /nccl-tests-2.17.1/build/all_reduce_perf
--- a/docs/software/container-engine/known-issue.md
+++ b/docs/software/container-engine/known-issue.md
@@ -117,7 +117,7 @@ When using the `cuda-dl` variant of the [AWS OFI NCCL hook][ref-ce-aws-ofi-hook]
 Dynamic linking makes the plugin interoperable across CUDA versions, but also reliant on a non-versioned symlink to the CUDA Runtime Library (e.g. `/usr/local/cuda/lib64/libcudart.so`).
 Such a link might not exist in some old container images, causing an error message at NCCL startup which mentions that `libcudart.so` could not be found.
 
-In these cases, we suggest the following paths to work around or solve the issue:
+In these cases, we suggest the following alternatives (pick one) to work around or solve the issue:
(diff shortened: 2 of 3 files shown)


docs/software/container-engine/known-issue.md
```diff
@@ -111,6 +111,31 @@ abc
 
 Notice the section `[annotations]` disabling Slurm and CXI hooks.
 
+## NCCL errors with a dynamically linked AWS OFI NCCL plugin: `libcudart.so could not be found`
+
+When using the `cuda-dl` variant of the [the AWS OFI NCCL hook][ref-ce-aws-ofi-hook] or `artifact` as [network stack source][ref-ce-netstack-source] (default in some vClusters), a dynamically linked AWS OFI NCCL plugin is mounted inside containers.
```
bcumming:
```suggestion
When using the `cuda-dl` variant of the [AWS OFI NCCL hook][ref-ce-aws-ofi-hook] or `artifact` as [network stack source][ref-ce-netstack-source] (default in some vClusters), a dynamically linked AWS OFI NCCL plugin is mounted inside containers.
```

docs/software/container-engine/known-issue.md
```diff
@@ -111,6 +111,31 @@ abc
...
+Dynamic linking makes the plugin interoperable across CUDA versions, but also reliant on a non-versioned symlink to the CUDA Runtime Library (e.g. `/usr/local/cuda/lib64/libcudart.so`).
+Such a link might not exist in some old container images, causing an error message at NCCL startup which mentions that `libcudart.so` could not be found.
+
+In these cases, we suggest the following paths to work around or solve the issue:
```
bcumming:
```suggestion
In these cases, we suggest the following alternatives (pick one) to work around or solve the issue:
```

bcumming (changes requested):

docs/software/communication/nccl.md
```diff
@@ -56,18 +56,18 @@ To use NCCL in a container, enable the [AWS OFI hook][ref-ce-aws-ofi-hook] in th
 
 ```toml
 [env]
-PMIX_MCA_psec="native" # (1)!
+PMIX_MCA_psec="^munge" # (1)!
```
msimberg, line 59: Just curious about this change: I've used `native` happily for quite a while. Why use `^munge` instead of `native`? Are there non-munge and non-native plugins that we want to allow?

bcumming (approved):

docs/software/container-engine/resource-hook.md
```diff
@@ -213,22 +213,28 @@ The [AWS OFI NCCL plugin](https://github.com/aws/aws-ofi-nccl) is a software ex
 Also see [NCCL][ref-communication-nccl] and [libfabric][ref-communication-libfabric] for more information on using the libraries on Alps.
 
 The Container Engine includes a hook program to inject the AWS OFI NCCL plugin in containers; since the plugin must also be compatible with the GPU programming software stack being used, the `com.hooks.aws_ofi_nccl.variant` annotation is used to specify a plugin variant suitable for a given container image.
-At the moment of writing, 4 plugin variants are configured: `cuda11`, `cuda12` (to be used on NVIDIA GPU nodes), `rocm5`, and `rocm6` (to be used on AMD GPU nodes alongside RCCL).
+At the moment of writing, the following plugin variants are configured:
+* For NVIDIA GPU nodes: `cuda12`, `cuda13`, `cuda-dl`.
+  The `cuda-dl` variant uses a plugin which is dynamically linked to CUDA, therefore being portable across versions, and is the generally recommended choice. Some issues may arise with old container images which don't provide generic symlinks to the CUDA Runtime (more details  [here]()).
```
msimberg:
```suggestion
  The `cuda-dl` variant uses a plugin which is dynamically linked to CUDA, therefore being portable across versions, and is the generally recommended choice. Some issues may arise with old container images which don't provide generic symlinks to the CUDA Runtime (more details [here]()).
```
but more importantly

```suggestion
  The `cuda-dl` variant uses a plugin which is dynamically linked to CUDA, therefore being portable across versions, and is the generally recommended choice. Some issues may arise with old container images which don't provide generic symlinks to the CUDA Runtime (more details  [here](where??)).
```

msimberg (dismissed):
CI is also complaining about a broken reference: https://github.com/eth-cscs/cscs-docs/actions/runs/27420041488/job/81042881905?pr=414#step:11:31.

docs/software/communication/nccl.md
Madeeks: I've been using `^munge` for a few months as well.
I can't remember exactly if I encountered a particular circumstance which prompted me to change, but a more general reasoning is avoiding to be tied to a specific implementation.
If I know that `munge` is the component giving problems, it feels more future-proof to me to disable it, rather than saying "only work with `native`": since I don't have visibility or control over Slurm or PMIx settings, I can't anticipate when or if another psec scheme becomes preferable in the future.
By just disabling munge, I leave Slurm the chance to fallback on anything that works.
At least, that is the intention.

docs/software/container-engine/resource-hook.md
Madeeks: Whoops, nice catch, forgot to link this to the Known Issue!

Madeeks: Should be fixed now

Madeeks:
Rebased and addressed the last remaining points from #413

bcumming (approved):

PR #415, opened 2026-06-15
Document SCALAPACK for non GPU-aware MPI

abussy:
Followup to #403. Since COSMA is the default PDGEMM library, and most CP2K calculations have at least some dense PDGEMM happening, simply setting `export MPICH_GPU_SUPPORT_ENABLED=0` is not quite enough.


RMeli (approved):

RMeli:
Good to be more explicit, thanks!

PR #417, opened 2026-06-16
Update known-issues.md

twrobinson:
The driver has been updated. Not sure this is the best place for known issues, so it might be best to remove the file and menu item.

### DIFF
--- a/docs/running/known-issues.md
+++ b/docs/running/known-issues.md
@@ -1,12 +1,5 @@
 [](){#ref-known-issues}
 # Known Issues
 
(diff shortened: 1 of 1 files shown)


bcumming:
I wonder whether these issues should be kept archived somewhere, at least for 6-12 months, given the propensity for these sorts of issues to pop up a while later?

twrobinson:
Agreed. For now you could add a top level menu called "Resolved" and put the item back. But I am still not sure it belongs in "Running jobs".

bcumming (approved):

PR #418, opened 2026-06-16
fix link to prohibited software

bcumming:
[SD-70148](https://jira.cscs.ch/browse/SD-70148)


twrobinson:
I would consider renaming the file from unsupported.md to prohibited.md.

bcumming:
The file name `$filename.md` is used to determine the final url:
```
docs.cscs.ch/software/$filename
```
Changing the file name changes the url, which will break links in the wild (i.e. if somebody has copied the url and saved it somewhere, it will become invalid).

In this case, probably not a big deal, because this is a very uneventful page, that likely isn't linked from anywhere.

Shall I do the rename?

twrobinson:
> The file name `$filename.md` is used to determine the final url:
> 
> ```
> docs.cscs.ch/software/$filename
> ```
> 
> Changing the file name changes the url, which will break links in the wild (i.e. if somebody has copied the url and saved it somewhere, it will become invalid).
> 
> In this case, probably not a big deal, because this is a very uneventful page, that likely isn't linked from anywhere.
> 
> Shall I do the rename?

I think it's better to rename it and risk a broken URL.

twrobinson (approved):

PR #420, opened 2026-06-16
Minor fixes to CE

Madeeks:
- Fix formatting of the AWS OFI NCCL hook section, so that variants are now correctly displayed as a bullet list, instead of a monolithic inline paragraph.
- Fixed incorrect formatting for reflink in AWS OFI NCCL hook section.
- Fix formatting of the numbered list in the dynamic AWS plugin known issue so that the annotation in the code block gets rendered correctly.


msimberg (approved):

msimberg:
Spell checker is on vacation (#422) so ignoring it for now.

PR #422, opened 2026-06-16
Remove spell checker

msimberg:
The spell-check action (https://github.com/check-spelling/check-spelling) was archived today (Jun 16th, 2026) and seems to have stopped working (unsure if because of the archiving or because of some other even correlated with the archiving). The action reports:
```
Run secpoll
Error: Found security advisory for version 0.0.26: 'Sorry'
Error: This is a fatal error ... asking Action Host to kill our workflow
Error: Process completed with exit code 200.
```
This seems to be a built-in check for the action itself (https://github.com/check-spelling/check-spelling/wiki/Feature:-Security-Polling).

Perhaps there will be more info on why the action was archived later and we can consider reenabling it through a fork, or just explore what other options there are. Until then I think the sanest (and safest) option is removing the checker entirely.


RMeli (approved):

msimberg:
Keeping it alive by ignoring the security check for now. Thanks @bcumming.

PR #427, opened 2026-06-17
Linkify libfabric manpage URL

msimberg:
The bare `https://...` doesn't become a clickable hyperlink. Change it to a proper hyperlink.


finkandreas:
Thank you, sorry we prepared first as slack message, where it was "working", and then decided to do it in the docs.

That one slipped through.

finkandreas (approved):

PR #431, opened 2026-06-22
removing storage LTS service doc

emdotef:


msimberg:
Fine by me to merge this but perhaps useful to ask: is it likely that a user will come around looking for it and wonder what happened to the service? Was it never in service (I haven't used it so I don't know the status of it)? Should one of the pages stay with a note saying the service no longer exists?

emdotef:
I would suggest too to leave a comment about the dismissed service. The fact is that the service is not available already since a while, during the years it was not accessible for a period then set accessible again and nobody claimed it. The service today has just one user that has not logged in for years.

We are promising a service that is no more there at least for the part where they had access to a UI where they could add new data. The only thing that is available is the resolution of the PID number through the handle service. 
If we continue to presume to have this service in place that is not, we will need then to find a solution that today we do not have and we do not have the resources to re-develop from the beginning the whole infrastructure.

bcumming:
Instead of removing the docs completely, I agree that they be updated to cover the service that we still offer.

If I understand correctly, new data can't be added, however there is some remaining support for people who have data in long term storage?

If that is the case, can we keep the page, with

* a notice at the top of the page saying that the service is no longer taking new customers, and is only providing access to data that was added before a certain date?
* docs on how to resolve PIDs through the handle service.

bcumming:
@emdotef do you have any feedback about the suggestion in my earlier comment?

emdotef:
Hi @bcumming,
I will come back with a new merge request and I will consider your suggestions.

Thanks
Monica

emdotef:
I will close this pull request and open a new one with new text

emdotef:
will provide a new pull request

PR #433, opened 2026-06-24
use internal links instead of external in CP2K docs

bcumming:
There were some links in the CP2K uenv docs that point to external docs when we have the same information internally: e.g. the Cray MPICH link pointed to NERSC docs, instead of the CSCS page.


RMeli (approved):
Good catch, thanks!

PR #437, opened 2026-06-30
Update the inference API page

msimberg:
Primarily updates the sections on creating an inference resource and API keys.

Also makes some minor adjustments to the page in general:
- added example response from models endpoint
- changed API key in examples to `$CSCS_INFERENCE_API_KEY` variable
- removed the "reducing token consumption" section
- other minor tweaks

### DIFF
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -9,7 +9,7 @@ Users consume tokens from a shared pool of models where requests are efficiently
 In order to maximize utilization and reduce costs, a reduced set of models is available. Because most of these models are trained by others, have inherent biases, and are aligned with their creators' principles, we highly recommend always auditing their results. Private model deployment is not supported.
 If you are interested to deploy a model that is not available in this service, we encourage using the [sml tool](https://github.com/swiss-ai/model-launch) developed by the Swiss AI community.
 
-Privacy and confidentiality are an essential to us. CSCS does not record user prompts or model responses, and your data does not leave the infrastructure we control. Nevertheless, including sensitive data in your prompts is not allowed. CSCS collects infrastructure metrics and telemetry, including prompt and response lengths, to monitor service quality.
+Privacy and confidentiality are essential to us. CSCS does not record user prompts or model responses, and your data does not leave the infrastructure we control. Nevertheless, including sensitive data in your prompts is not allowed. CSCS collects infrastructure metrics and telemetry, including prompt and response lengths, to monitor service quality.
 
 
 ## Service at a glance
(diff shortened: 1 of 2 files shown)


naevtamarkus:
I think the changes look great!

docs/services/inference/api.md
```diff
@@ -2,35 +2,14 @@
...
-
-Private model deployment is not supported.
+In order to maximize utilization and reduce costs, a reduced set of models is available. Private model deployment is not supported.
 If you are interested to deploy a model that is not available in this service, we encourage using the [sml tool](https://github.com/swiss-ai/model-launch) developed by the Swiss AI community.
 
-Usage of sensitive or personal data is not allowed.
-For privacy reasons, CSCS does not track user prompts or model responses.
-However, CSCS collects infrastructure metrics and telemetry, including prompt and response lengths, in order to monitor the service quality.
-
+Your privacy and the confidentiality of your data are paramount to us. CSCS does not track user prompts or model responses, and your data does not leave the infrastructure we control. Nevertheless, including sensitive or personal data in your prompts is not allowed. CSCS collects infrastructure metrics and telemetry, including prompt and response lengths, to monitor service quality. Because these models are trained by others, have inherent biases, and are aligned with their creators' principles, we highly recommend always auditing their results.
```
msimberg: About

> Usage of sensitive or personal data is not allowed

Can we actually say this? What does "is not allowed" imply? Can I send a prompt with "My name is foo and I'm 42 years old". That's certainly personal data, but why is it not allowed? More concretely: should we just stick to "sensitive" rather than "sensitive and personal"?

Do we need to mention that inputs can be cached, which implies _some_ level of data retention?

bcumming: I agree that this is vague.

Are you saying that data data that is protected by law (e.g. medical data) should not be sent to the API?

Furthermore, this feels like three separate points

* sensitve data
* CSCS record logs
* don't blindly trust the results.

So we might want to have an empty line between each sentence to make three separate paragraphs to better indicate that these are separate concerns.

naevtamarkus: There is a provision in the ETH BOT (which applies to all CSCS users) that says:
When processing personal data, the legal requirements of data protection must be complied with. In particular, Art. 36a-36f of the ETH Act apply. If you want to see what's considered "personal sensitive" you can check page 22 here: https://ethz.ch/content/dam/ethz/associates/services/Anstellung-Arbeiten/Downloads/files/rechtliches/ETH_Act.pdf

This service is for CSCS users, so agreeing to this is a given... but still, I think it's best to mention it. 

Name and age are not there, but nationality or date of birth is.
My suspicion is that's best to leave "personal or sensitive" as is.
As for splitting it in three pints, it feels like making it more important than it is...

naevtamarkus (approved):
Looks great, possible improvements on the privacy aspects, to be fixed on a later PR

PR #446, opened 2026-07-10
Polishing inference docs

msimberg:
I made apertus it's own card in the service at a glance, it's a bit more visible this way. It's also a "foundation" with a "pillar" icon and sits at the bottom of the list, so too many metaphors to even count.

I reworked the intro a bit to hopefully flow a bit better. The first paragraph focuses on the service and the models, and mentioning the sml tool for self hosted models. The next paragraph focuses on not recording prompts and collecting infrastructure metrics.

I removed the explicit "sensitive data not allowed" and link to the policies instead in a separate paragraph.

I added another explicit link to the coding agents guide and setting up coding agents with the inference service to the intro as well to highlight it a bit better.

Some of these are definitely subjective changes, so I welcome feedback. Some things I might have gone too far, and some not far enough. Let me know what you think.

### DIFF
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -15,6 +15,9 @@ However, CSCS collects infrastructure metrics and telemetry, including prompt an
 
 The [CSCS policies][ref-policies] apply to users of the inference service.
 
+For use of coding agents on and around Alps, see the [coding agents on Alps guide][ref-coding-agents].
+The inference API [can be used with most coding agents][ref-inference-api-coding-agents-setup].
+
 <div class="grid cards" markdown>
 -   <p style="text-align:center">Visit <a href="https://inference.status.cscs.ch/">inference.status.cscs.ch</a> for the status of the inference service, models, and latest announcements.</p>
 </div>
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -1,8 +1,6 @@
 [](){#ref-inference-api}
 # LLM Inference API Service
 
-[](){#ref-inference-api-beta}
-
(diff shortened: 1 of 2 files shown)


bcumming (approved):
Looks good.

Still need more information about exactly what information is collected and policies, but that is for somebody else (not me) to decide.

docs/services/inference/api.md
```diff
@@ -200,6 +213,12 @@ The following OpenAI- and Anthropic-compatible endpoints are available.
 When using the endpoints for example [through agents][ref-inference-api-coding-agents-setup], the framework will handle API requests for you.
 For information on how to use the endpoints directly, see the [OpenAI](https://developers.openai.com/api/reference/overview) and [Anthropic](https://platform.claude.com/docs/en/api/overview) documentation.
 
+[](){#ref-inference-api-available-models}
+## Available models and pricing
+
+Available models, along with pricing information, are listed on the [cscs2go Inference API page](https://2go.cscs.ch/offering/swiss_academia/#inference-api).
+The available models can also be listed for a given API key using the [`models` endpoint][ref-inference-api-endpoints] or on the Inference API UI page when creating a new key.
+
```
msimberg: This will still need elaboration, since it'll be possible to pay for inference using compute hours from an allocation.

naevtamarkus:
Hi Mikael, I don't really like much the way the top part is now. The reader's flow breaks often, going from detail to high level to detail again, and from warning to main text and so fort.

I think we should keep the structure:
- First the high level stuff. What this service is, its core advantages (inc. the service "at glance")
- Then the policies, alternatives. 
- Then the quick-start guide, examples
- And finally the caveats

I'm not fully convinced about the caveats part being at the bottom, maybe after policies before examples?

msimberg:
Closing in favour of #451 and #448.

PR #448, opened 2026-07-13
Inference API pricing and status page links

msimberg:
Splitting out the pricing, models, and status page links from #446.


naevtamarkus (approved):
Perfect, thanks!

PR #458, opened 2026-07-15
[fix] change name to Libfabric

toxa81:
In the current documentation, [Cray MPICH](https://docs.cscs.ch/software/communication/libfabric/#libfabric) is pointing to Libfabric section


msimberg (approved):
Good catch, thanks!

PR #466, opened 2026-07-30
Add a workaround to the image build document.

gwangmu:
Resolves: https://jira.cscs.ch/browse/VCUE-1524

Since a security mitigation against CVE-2026-53264 on 30.07.2026, it was reported that the image build service using Podman stopped working properly on some clusters. This MR adds a workaround about how to mitigate the issue.


gwangmu:
The original mitigation that required this workaround is gone on most clusters as of this writing, but I think it's harmless to leave a short trick that users can try in case something comes up in the future.

bcumming (approved):

PR #472, opened 2026-08-19
[SANTIS] Update documentation

kraushm:
This should be merged on 26 August, when most new features described here will be deployed.


docs/clusters/santis.md
```diff
@@ -1,22 +1,26 @@
...
+The number of nodes can vary as nodes are added or removed from other clusters on Alps.
+See the [Slurm documentation][ref-slurm-partitions-nodecount] for information on how to check the number of nodes.
 
-There are four login nodes, labelled `santis-ln00[1-4]`.
-You will be assigned to one of the four login nodes when you ssh onto the system, from where you can edit files, compile applications and start simulation jobs.
+There are two login nodes, `santis-ln00[1,2]`, which are repurposed compute nodes from the pool of 302 GH200 nodes.
+You will be assigned to one of the two login nodes when you ssh onto the system, from where you can edit files, compile applications and launch batch jobs.
+This leaves 300 nodes available for compute jobs.
+
+Each GH200 node has four Grace-Hopper chips, giving 288 CPU cores, 4 GPUs and approximately 870 GB of RAM per node.
```
msimberg: Does this need to be repeated from the ref-alps-gh200-node link?

msimberg: And the 870 GB of RAM can be a bit misleading... since half of that is GPU memory and behaves quite differently. Same comment applies further down with the mention of ~217 GB of memory per module.

docs/clusters/santis.md
```diff
@@ -64,59 +99,104 @@ In addition to the climate and weather uenv, all of the
...
+```
+
+If no GPU is requested, the job may receive only a minimal CPU allocation and will not run on an exclusive node.
+
+#### Multi-GPU jobs with P2P/IPC
+
+Because device isolation is enforced via cgroups, multi-GPU jobs that rely on GPU P2P/IPC (for example MPI across GPUs on the same node) must add:
+
+```bash
+--gres-flags=allow-task-sharing
```
msimberg, line 144: I guess we're not aware of a way to make this the default when a full node is allocated?

docs/clusters/santis.md
```diff
@@ -64,59 +99,104 @@ In addition to the climate and weather uenv, all of the
...
+```
+
+This keeps all GPUs allocated to a job visible to all tasks of that job while preserving per-task `CUDA_VISIBLE_DEVICES` bindings. Without this flag, intra-node GPU communication may fail.
```
msimberg:
```suggestion
This keeps all GPUs allocated to a job visible to all tasks of that job while preserving per-task `CUDA_VISIBLE_DEVICES` bindings. Without this flag, intra-node GPU-GPU communication will fail.
```

msimberg (commented):
Just general comments at this point (especially since it's not for merging yet). What's the motivation for duplicating a lot of the information that is elsewhere? E.g. GH200 info, or checking quota. Not completely against it, but you must have had some reason for adding it instead of linking to one canonical place in the docs.

I guess the important part of these changes is the introduction of the shared partition (officially) and the related impact on GPU-GPU communication?

docs/clusters/santis.md
kraushm: No, not yet. Also, I need to specify that this is an `srun` option and cannot simply be added to the #SBATCH block.

docs/clusters/santis.md
kraushm: Sorry, my bad. With the link above there is no need for this line. Will remove it.

msimberg (approved):
Looks good to me in general, but have not read everything in detail.

bcumming (approved):
Some of the information about checking quota and the  Store and Scratch filesystems replicates documentation elsewhere, and could be simplified later.

PR #480, opened 2026-08-26
Replace MkDocs with Zensical

bcumming:
The robust error detection (the `--strict` flag) can now be used reliably in CI/CD to detect and prevent broken docs from being deployed.

This PR will require care to merge, in order to ensure that CI/CD continues to work for other pending PRs.

Note:
* the [PR preview](https://cscs-docs-preview.svc.cscs.ch/480) is rendered using Zensical
* we will need to retain the `mkdocs.yaml` configuration for the first deployment of this merged change to deploy correctly


finkandreas (approved):
lgtm

.github/workflows/docs.yaml
```diff
@@ -18,13 +18,53 @@ jobs:
...
+          # stop GitHub Pages from running the output through Jekyll
+          touch gh-pages/.nojekyll
+          git -C gh-pages add --all
+
+          if git -C gh-pages diff --cached --quiet; then
+            echo "site is unchanged, nothing to publish"
+          else
+            git -C gh-pages commit -m "Deployed ${GITHUB_SHA} with zensical"
+            git -C gh-pages push origin gh-pages
+          fi
```
msimberg, line 70: This feels a bit unnecessarily manual, but I've no objection to keeping this version and optionally changing this later.

Just FYI what I mean by it being unnecessarily manual is that it's possible to publish github pages without the gh-pages branch: https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow. We publish e.g. pika's documentation with this workflow: https://github.com/pika-org/pika/blob/947cf61a14fcafc4d472fb3fcddf3bd2e89d792e/.github/workflows/docs.yml#L48-L73.

.github/workflows/docs.yaml
```diff
@@ -18,13 +18,53 @@ jobs:
...
+      - run: zensical --version
```
msimberg: Also unrelated to your changes, so optional and a nitpick: move this to be the first thing that runs so that if the deploy step fails you know what version was actually used?

docs/build-install/uenv.md
```diff
@@ -281,13 +281,10 @@ To use them, add them to the module environment:
     The generation of modules can be customised by editing the `<build-path>/config/user/modules.yaml` file _before_ running `build`.
     See the [Spack modules] documentation.
 
-[Chaining Spack Installations]: https://spack.readthedocs.io/en/latest/chain.html
```
msimberg, line 284: Are these removed because zensical warns about them being unused or are these drive-by changes?

zensical.toml
```diff
@@ -0,0 +1,330 @@
...
+# error that fails the build.
+# Unresolved autorefs are always reported, and are not configurable here.
+[project.validation]
+# a link to a page that does not exist
+invalid_links = true
+# a link to an anchor that does not exist on the target page
+invalid_link_anchors = true
+# must stay disabled: every [text][ref-...] link in the docs is an autoref, not
+# a Markdown reference link, so this flags all ~1000 of them
+unresolved_references = false
```
msimberg, line 242: This is a bit sad. Do you know if this is something they're planning to support? I think mkdocs warned about incorrect autorefs, now?

zensical.toml
```diff
@@ -0,0 +1,330 @@
...
+accent  = "blue"
+toggle  = {icon = "material/brightness-7", name = "Switch to dark mode"}
+
+# Palette toggle for dark mode
+[[project.theme.palette]]
+media   = "(prefers-color-scheme: dark)"
+scheme  = "slate"
+primary = "white"
+accent  = "blue"
+toggle  = {icon = "material/brightness-4", name = "Switch to light mode"}
```
msimberg: One of the toggle texts is probably wrong, no? I guess they're cycled in the order, so probably the last one should be "Switch to auto mode"? Might already have been wrong before.

msimberg (commented):
Looks great to me, thanks @bcumming for pushing this.

I have a few nitpicks that are not super blocking and can be postponed.

This is possibly just an old-man comment: the new search works great, _but_ the text is quite a bit smaller in the search results compared to the old search. To me, a bit unnecessarily small, but I know this is likely very subjective.

Adding this as "comment", but please take it as an "approve" if you think these are better resolved separately or not at all.

.github/workflows/docs.yaml
RMeli: Yeah, it looks quite complicated... On my personal notes website, which I also moved to zensical, I use [this](https://zensical.org/docs/publish-your-site/#github-pages), which uses some of the actions referred in the GitHub documentation that Mikael linked.

jpcoles-cscs:
The blurred background and search box have almost no contrast. It's hard to see there’s a new box, especially in dark mode. See image.
image (29)

.github/workflows/docs.yaml
bcumming: Thanks.
I will look into the alternatives - I will admit to using an agent to write that.

zensical.toml
bcumming: Broken link detection was a hard requirement for the pull request.
That option gives false positives for `autorefs` links, which is what we use.
Broken autorefs are handled correctly by the options above that one in the config.

bcumming:
> The blurred background and search box have almost no contrast. It's hard to see there’s a new box, especially in dark mode. See image. image (29)

There is no way to tune opacity, border properties or font size in the search box.

I have raised this with the developers.

.github/workflows/docs.yaml
bcumming: This mirrors the current approach of deploying via a branch, while using Actions also requires a GitHub settings change.

The workflow isn't too complicated, and it has the benefit of being inline and not hidden behind a pre-packaged action. So I am on the fence about the pros and cons of either approach.

Let's leave it as is for now, and visit it once the deployment is successful.

docs/build-install/uenv.md
bcumming: yes: the new checker catches more categories of link stink.

zensical.toml
bcumming: Indeed. I have fixed.

PR #486, opened 2026-09-01
Improve zensical search styling

bcumming:
The zensical search box comes with some iffy default styling:

- small fonts
- low contrast making text and regions difficult to parse visually
- clicking the search button takes focus to a different part of the screen, but it is not easy to see where you should move your focus

Zensical does not net provide hooks for customizing search look and feel, so this implements a temporary fix based on inspecting the internals of the search tool.
- it might break if the internal search tool implementation changes.

### DIFF
--- a/docs/javascripts/search-styling.js
+++ b/docs/javascripts/search-styling.js
@@ -1,0 +1,209 @@
+/*
+ * Stop-gap styling for the search overlay.
+ *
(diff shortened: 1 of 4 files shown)


msimberg (approved):
Looks nicer to me, thanks for investigating this! Up to you if the extra js/css is worth the maintenance effort (I suppose it won't be touched for a very long time), but I suppose if/when zensical implement a nicer way to customize that this is acceptable.

PR #491, opened 2026-09-04
Provide a link to external ICON docs instead of removing them entirely

havogt:


msimberg (approved):
Also good, thank you @havogt.

PR #495, opened 2026-09-08
Add VS Code Copilot setup for the inference API

PhilHippo:
Add instructions for using the inference API from VS Code Copilot Chat

### DIFF
--- a/docs/services/inference/api.md
+++ b/docs/services/inference/api.md
@@ -276,7 +276,7 @@ For information on how to use the endpoints directly, see the [OpenAI](https://d
 [](){#ref-inference-api-coding-agents-setup}
 ## Setting up coding agents to use the inference service
 
-Below are instructions for setting up [Claude Code](https://claude.com/product/claude-code) and [OpenCode](https://opencode.ai) to use the inference service.
+Below are instructions for setting up [Claude Code](https://claude.com/product/claude-code), [OpenCode](https://opencode.ai) and [VS Code Copilot](https://code.visualstudio.com/docs/copilot/overview) to use the inference service.
 For more information on using coding agents on Alps, see the [coding agents guide][ref-coding-agents].
 
 See the [available models table][ref-inference-api-available-models] for context sizes.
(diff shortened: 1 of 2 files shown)


docs/services/inference/api.md
```diff
@@ -377,6 +377,82 @@ Once configured, you can choose models configured in the config with `/models` o
...
+- `name` is the display name shown in the model picker,
+- `url` is the base URL of the inference API, to which VS Code appends `/chat/completions`,
+- `toolCalling` enables tool use, which is required for agent mode,
+- `vision` enables image input and should only be set for multimodal models,
+- `maxInputTokens` is the maximum context length of the model from the [available models table][ref-inference-api-available-models], and
+- `maxOutputTokens` is the maximum number of tokens in a single response, for which `16384` is a good default for coding tasks.
+
+!!! note
+    VS Code treats the sum of `maxInputTokens` and `maxOutputTokens` as the context window of the model.
+    If you encounter context length errors, reduce `maxInputTokens` by the value of `maxOutputTokens`.
```
msimberg: If this is the case, why not just reduce the maxOutputTokens from maxInputTokens in the example? I think it makes more sense than having this warning (the second sentence; the first one is still useful).

docs/services/inference/api.md
```diff
@@ -377,6 +377,82 @@ Once configured, you can choose models configured in the config with `/models` o
...
+??? info "Maximum output tokens per model"
+    The inference API does not publish a per-model output limit, and public providers of the same open-weight models use different values.
+    Values reported by model vendors and public providers are:
+
+    - `moonshotai/Kimi-K2.7-Code`: 32,768 by default on the Moonshot platform,
+    - `zai-org/GLM-5.2`: up to 131,072,
+    - `google/gemma-4-31B-it` and `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16`: 16,384 on most public provider,
+    -  Apertus models: no published limit.
+
+    Requests that exceed the limit enforced by the inference API fail with an error, in which case lower `maxOutputTokens`.
```
msimberg: I would likely remove this, or move this to the top level "## Setting up coding agents to use the inference service" section if we keep it. As far as I know we don't set a max output limit and in that case I don't think we should be referring to limits from other providers.

Checking with NETC if we actually do have limits.

docs/services/inference/api.md
```diff
@@ -377,6 +377,82 @@ Once configured, you can choose models configured in the config with `/models` o
...
+- Click "Add Models" in the top right corner and select "Custom Endpoint".
+- Enter a name for the group of models, for example `CSCS Inference`.
+- Enter your API key when prompted for the `apiKey` value.
+  VS Code stores the key in its secret storage and only references it from the configuration file.
+- Select "Chat Completions" as the API type.
+
+VS Code then opens the `chatLanguageModels.json` configuration file with an entry for the new provider.
+Add the models that you want to use to the `models` list, leaving the generated `apiKey` reference untouched:
+
+- After filling the json , save and close the file.
```
msimberg:
```suggestion
- After filling the json, save and close the file.
```

msimberg (changes requested):
Thanks @PhilHippo! A couple of minor comments. Need to check the output tokens limit...

docs/services/inference/api.md
msimberg: According to David we set no limits on the outputs, so the practical limit is the context size. I'd probably drop this? The main reason to set it is to encourage VSCode to compact before it hits the limit, but that would hopefully be best done with a separate setting (OpenCode and pi allow setting a percentage).

PhilHippo:
Thanks for the review. I applied the changes! I double checked the `maxInputTokens` and `maxOutputTokens` fact.

msimberg (approved):

